PR #227, opened 2020-11-08
popular user discovery update according to new hivemind.

talhasch:
@feruzm note that `hiveUri` in `src/config.ts` must be updated along with this update.


feruzm (approved):
lgtm

PR #242, opened 2020-11-15
Witnesses page

talhasch:


feruzm (commented):
Looking good, only in desktop view, I would prefer bit smaller screen width.
Right now, it is from edge of one side to other edge of another side. Perhaps make it smaller on desktop view same size as post content area size, other sides can stay white space.

image

talhasch:
@feruzm we can limit size for large screen for both web and desktop?

feruzm:
@talhasch yes, some padding on desktop wouldn't hurt as well. Default screen size for desktop app is smaller than browser default size, at least for some users, desktop could have less padding compare to browser, imo.

talhasch:
ok. have set max-width: 1200px. looking good for both web and desktop.

feruzm (approved):
lgtm

PR #259, opened 2020-11-24
hot topics

talhasch:
Adds hot topics section above trending tags list which we create a list out of a defined tag list.


talhasch:
@feruzm i am open for color suggestions

feruzm:
@talhasch let's keep icon same color as font, just like in dark mode, they don't have to be too much attention attracting.

feruzm:
Please add `Gaming` section with `hive-140217` as well.

talhasch:
done

feruzm (approved):
lgtm 🚀

PR #260, opened 2020-11-25
renderer version update

talhasch:


talhasch:
added setting path to renderer

feruzm (approved):

PR #279, opened 2020-12-07
Ecency wallet design update

talhasch:
Similar design as hive wallet


talhasch:
@feruzm updated ecency wallet

feruzm:
few minor changes in width of token type buttons and icon, Buy Points could be turned to button as well. 
We should be ok for release after these...

feruzm:
@talhasch Great, only making Buy Points blue button and moving it slightly up, will be perfect.

image

talhasch:
@feruzm done

feruzm (approved):
lgtm

PR #281, opened 2020-12-07
Before release

talhasch:


src/common/components/friends/index.tsx
```diff
@@ -140,13 +140,29 @@ export class List extends Component<ListProps, ListState> {
...
+        const {account, mode} = this.props;
+
+        const follower = (mode === "following" ? account.name : search);
+        const following = (mode === "following" ? search : account.name);
+
+        let connection: boolean;
+        let ac: Account | null = null;
         try {
-            data = await this.fetch(search.replace("@", ""), 1);
+            const resp = await getRelationshipBetweenAccounts(follower, following);
```
feruzm, line 151: Couldn't we just do live search from data[], if not all data[] is loaded, Load more button could stay and by keeping search input we can do search once load more clicked and search from new data[]?

talhasch: i don't get purpose of this?

feruzm: Search requiring user to press `enter` and should use exact username, so search looses its purpose somehow. Instead we can do `filter` on array `data[]`. If user didn't load all followers, then `Load more` could keep showing and search continues to filter from what data is available already in memory.

talhasch:
@feruzm implemented api endpoints for friend search

talhasch:
@feruzm added typeahead search. pls check it.

talhasch:
@feruzm pushed workaround for pin/unpin

feruzm (approved):

PR #286, opened 2020-12-10
Transfer dialog improvements & delegation.

talhasch:
@feruzm Please do not merge this PR now. It is better to merge once all missing wallet features done.

Defined different success messages per mode. Please review/check them in locale file.


src/common/components/delegated-vesting/index.tsx
```diff
@@ -24,149 +33,147 @@ import parseAsset from "../../helper/parse-asset";
...
+                                },
+                                onKey: (key) => {
+                                    this.stateSet({inProgress: true});
+                                    delegateVestingShares(activeUser.username, key, username, "0.000000 VESTS")
+                                        .then(() => this.fetch())
+                                        .catch(err => error(formatError(err)))
+                                        .finally(() => this.stateSet({inProgress: false}))
+                                },
+                                onHot: () => {
+                                    delegateVestingSharesHot(activeUser.username, username, "0.000000");
```
feruzm, line 132: I believe `vesting_shares` should be in `0.000000 VESTS`, just numbers wouldn't work.

src/common/components/transfer/index.tsx
```diff
@@ -377,26 +420,29 @@ export class Transfer extends Component<Props, State> {
...
+            case "withdraw-saving":
+                transferFromSavingsHot(username, to, fullAmount, memo);
                 break;
-            case 'power-up':
-                prms = transferToVestingHot(username, to, fullAmount);
+            case "power-up":
+                transferToVestingHot(username, to, fullAmount);
+                break;
+            case "delegate":
+                const vests = this.hpToVests(Number(amount), false);
```
feruzm, line 444: `withSymbol` should be true otherwise we would get serialization issue.

feruzm (changes requested):

src/common/components/transfer/index.tsx
talhasch: hive signer doesn't accept with amount with `VESTS` that's why withSymbol exists in that function.

feruzm: @talhasch Really, I tried without VESTS with hivesigner, it said serialization issue.

feruzm: here screenshot of error: 
image

talhasch: ok let me check it

talhasch:
@feruzm can you check power down?

feruzm:
@talhasch ok

feruzm:
@talhasch so far my comments

- Delegation shouldn't allow to own profile
image

- Power down looks good and simple, but we need setting route as well like we used to have.

talhasch:
@feruzm  added withdraw routes into the dropdown on hive power section on wallet page.

Delegation used to set active user as target user by default. Fixed that. If user wants to delegate themself anyway, blockchain rejects.

Doing some more checks. Then ready to merge this.

talhasch:
@feruzm ready to merge

feruzm (approved):
lgtm, good call on cleaning up pre-hardfork variables.

PR #291, opened 2020-12-16
Voters modal update

talhasch:
Updated voters list with to generic user list mixin. 
Created a custom pagination component.
Deleted `react-bootstrap-table-next` and `react-bootstrap-table2-paginator` dependencies. Saves us 25KB at bundle size.


feruzm (approved):
nice, looking good... this should fix modal overlay issue on mobile devices also right?!

talhasch:
yep

PR #298, opened 2020-12-17
Community select

talhasch:
Adds community select to editor page.

### DIFF
--- a/src/common/api/hive.ts
+++ b/src/common/api/hive.ts
@@ -6,6 +6,8 @@ import {FullAccount, AccountProfile, AccountFollowStats} from "../store/accounts
 
 import parseAsset from "../helper/parse-asset";
 
+import isCommunity from "../helper/is-community";
+
 import SERVERS from "../constants/servers.json";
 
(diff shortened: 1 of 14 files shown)


feruzm (commented):
Looks good to me, simple and effective.
Could we also add (if user posting in community, add reblog operation into same transaction) ?
Right now, if user posts in community, they will have to manually reblog so it shows on their blog as well, but that's unnecessary step and adding option to editor like (add to my blog as well) also unnecessary. Just making it easy for user, default reblog their new post from community into their blog as well while posting.

talhasch:
> Looks good to me, simple and effective.
> Could we also add (if user posting in community, add reblog operation into same transaction) ?
> Right now, if user posts in community, they will have to manually reblog so it shows on their blog as well, but that's unnecessary step and adding option to editor like (add to my blog as well) also unnecessary. Just making it easy for user, default reblog their new post from community into their blog as well while posting.

ok

talhasch:
@feruzm pushed auto reblog for community posts

feruzm (approved):

PR #302, opened 2020-12-18
Mobile improvements

talhasch:
Pls do not merge now.

Updated entry list item styles. It has to be tested in detail for mobile view and grid view on desktop sizes.


feruzm:
Looking good... 

image
let's change reblog button bottom of page to `...` context menu and inside add Reblog button. On mobile app we have vertical dots for context menu. That way we can add features like Bookmark, Pin right from feed. Text of Reblog can be `[icon] esteemapp reblogged` click on this could open esteemapp profile page.

image
Promoted should probably go under everything like twitter, or make it bit less noticeable without blue background just text similar to reblog.

feruzm:
image

also would be good to align cover image in desktop grid view

talhasch:
> Looking good...
> 
> image
> let's change reblog button bottom of page to `...` context menu and inside add Reblog button. On mobile app we have vertical dots for context menu. That way we can add features like Bookmark, Pin right from feed. Text of Reblog can be `[icon] esteemapp reblogged` click on this could open esteemapp profile page.
> 
> image
> Promoted should probably go under everything like twitter, or make it bit less noticeable without blue background just text similar to reblog.

putting those controls bottom of element needs lots of effort and doesn't sounds as good approach. it may needs to create duplicates of those element one for mobile (will be bottom) and one for larger devices (will be top).

talhasch:
> image
> 
> also would be good to align cover image in desktop grid view

ok i

talhasch:
> image
> 
> also would be good to align cover image in desktop grid view

ok will give a try for this.

talhasch:
@feruzm pls check recent additions.

feruzm:
@talhasch language selection and Faq locations looks good, but better if Downloads is above Market price.

talhasch:
> @talhasch language selection and Faq locations looks good, but better if Downloads is above Market price.

it doesn't look good imo 
Ekran Resmi 2020-12-18 18 11 27

feruzm:
@talhasch ok leave it as is then, bottom of market price.

talhasch:
@feruzm i think this is ready to merge

PR #307, opened 2020-12-21
Visual fixes for mobile screens

talhasch:
@feruzm  If we take theme and language button under search box there will be too much visible buttons for new users. 

Not a good UX imo. Better to leave them up there.

Ekran Resmi 2020-12-21 15 52 28

### DIFF
--- a/src/common/components/intro/_index.scss
+++ b/src/common/components/intro/_index.scss
@@ -48,7 +48,7 @@
     background-color: $dark-grey-blue;
     color: $silver;
   }
-  height: 180px;
+  height: 200px;
 
   @media (min-width: $md-break) {
     height: 240px;
--- a/src/common/components/navbar/_index.scss
+++ b/src/common/components/navbar/_index.scss
@@ -272,6 +272,7 @@
 
     .switch-language {
       left: 60px;
+      top: 20px;
 
       @media (min-width: $md-break) {
         left: auto;
--- a/src/common/components/navbar/_index.scss
+++ b/src/common/components/navbar/_index.scss
@@ -233,7 +233,6 @@
 
       @media (min-width: $xxl-break) {
         width: 500px;
-        margin-right: 40px;
       }
     }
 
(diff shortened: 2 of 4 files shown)


feruzm:
@talhasch I just look at symmetry... 🤔 maybe move write button same line as login and signup?
and we can try swapping logo and theme/languages buttons, logo on left side and then follows this icons? How would that look?

feruzm:
@talhasch Check this out, I think better management of space and we will have better alignment. what do you think?
image

feruzm:
@fildunsky Could you check this and suggest us some variation?

talhasch:
@feruzm let me try to implement with code

talhasch:
Ekran Resmi 2020-12-21 16 25 41

@feruzm ?

feruzm:
@talhasch yes for now we can use that until Fil suggest something better/different. Only can we add padding to logo from left so it will alight with edge of search field ?

talhasch:
ok

talhasch:
@feruzm pls check latest update

feruzm (approved):
lgtm

fildunsky:
suggest

I am sorry for delay, here some of my thoughts. Feel free to ignore them if you disagree :)
1. I would make language button same color as theme button because it is not the most prioritized thing user should focus on.
2. Just found out that I read `Discover Communities` as one thing and maybe it will be confusing for beginners to find out that those are two different links. What if we divide them using light-gray ` | ` symbol? What do you think, guys?
3. Also I am not sure but I've seen many popular login screens and usually `Login` is looking like a link and `Signup` is active button.
4. By the way, why there is a pencil icon `Write a post` if user is not logged in?

talhasch:
@fildunsky 1,2 and 3 are good suggestions. i think we should implement. what do you say @feruzm 

The reason of 4 is to allow new visitors to see editor page and start writing easily. Once they click publish button they see login dialog. So they can login or signup and continue publishing.

feruzm:
Good suggestions, 

for 3. I would say separator doesn’t necessarily convey separation. Maybe more subtle is to underline those links on mobile so it should be more visible without extra character?

fildunsky:
`underline those links on mobile`
if we will underline those links on mobile then logically we should underline all the links in the UI. Not sure if that will look clean... Maybe it's better to keep it as it is for now.

PR #322, opened 2021-01-03
vp to user nav menu

talhasch:


feruzm (changes requested):
In night mode, color slightly not visible... also let's make it like `Voting Power ^ 34.5%` so later we might add downvote power next to it...
image

talhasch:
@feruzm pls review latest updates

feruzm (approved):
lgtm

PR #342, opened 2021-01-20
Font color change

feruzm:


talhasch (approved):
changed color variable name. added themify to use new color only in day theme.

PR #346, opened 2021-01-23
Password update

talhasch:


talhasch:
@feruzm ?

feruzm (approved):
Looks good, only this gives me thought on how to show/fix hivesigner that it detects if account has changed password that is saved in localstorage

talhasch:
> Looks good, only this gives me thought on how to show/fix hivesigner that it detects if account has changed password that is saved in localstorage

Comparing public keys may be a solution. Hivesigner creates public keys from password that stored in local storage and compares with one of the actual public keys. Does it makes sense?

talhasch:
@feruzm do we need to refresh token after password update?

feruzm:
Yea that’s what I thought of doing as well until our reimplementation will be ready.

Access token should continue to work as long as posting authority is not removed in process. We might need to update active key in redux if it is kept.

PR #364, opened 2021-02-10
Tipping

talhasch:


feruzm:
looking great... couple minor things if we can add

- change icon for amount
image

- `Tip for username/permlink` add `@` front of username so in wallet later we can linkify them

talhasch:
> looking great... couple minor things if we can add
> 
> * change icon for amount
>   image
> * `Tip for username/permlink` add `@` front of username so in wallet later we can linkify them

@feruzm done

feruzm (approved):
lgtm

PR #370, opened 2021-02-17
Community rewards

talhasch:
@feruzm pls check for typo with new copy


feruzm (approved):
lgtm, I would change memo from backend change to `Rewards for @username/permlink` or just `@username/permlink` clickable link

PR #384, opened 2021-02-23
human readable chain error feedback

talhasch:


src/common/api/operations.ts
```diff
@@ -46,6 +48,21 @@ export interface CommentOptions {
...
+        return _t("chain-error.min-root-comment");
+    } else if (/Your current vote on this comment is identical/.test(strErr)) {
+        return _t("chain-error.identical-vote");
+    } else if (/Please wait to transact, or power up/.test(strErr)) {
+        return _t("chain-error.insufficient-resource");
+    } else if (/Cannot delete a comment with net positive/.test(strErr)) {
+        return _t("chain-error.delete-comment-with-vote");
+    } else if(/comment_cashout/.test(strErr)){
+        return _t("chain-error.comment-cashout");
+    }
```
feruzm, line 64: There is also error on voting old posts, would be good to add that as well. @talhasch

feruzm (approved):
lgtm 👍

PR #394, opened 2021-03-05
Auto scroll to comment

talhasch:


feruzm:
@talhasch seem to work great, we only need to check if render helper can link those comment linking content properly?!

talhasch:
as long as post page receive a comment link it will scroll to it. so we can merge this and check render-helper separately.

feruzm (approved):
lgtm

PR #398, opened 2021-03-09
i18n html sanitization

talhasch:


feruzm (approved):
lgtm, good call on xss filter

PR #410, opened 2021-03-19
Auto fetch latest surfer version

talhasch:


src/common/components/download-trigger/index.tsx
```diff
@@ -2,10 +2,31 @@ import React, {Component, Fragment} from 'react';
...
 
-export class DialogContent extends Component {
+interface ContentState {
+    desktopTag: string;
+}
+
+export class DialogContent extends BaseComponent<{}, ContentState> {
+
+    state: ContentState = {
+        desktopTag: "3.0.14"
```
feruzm, line 18: 3.0.15 should be default because of api path change. if github api had issues, user will get old not working version.

talhasch: done

PR #414, opened 2021-03-22
Reblog delete

talhasch:


src/common/i18n/locales/en-US.json
```diff
@@ -147,7 +147,11 @@
     "reblog": "Reblog",
     "confirm-title": "Reblog to @{{n}}'s blog?",
     "confirm-ok": "Reblog",
-    "success": "Reblogged!"
+    "success": "Reblogged!",
+    "delete-reblog": "Delete Reblog",
```
feruzm, line 151: Let's change strings to these `Undo Reblog`, `Undo reblog?`, `Undo`,  `Reblog removed`

talhasch:
Done

Feruz M ***@***.******@***.***>> şunları yazdı (22 Mar 2021 15:17):

Reblog removed

src/common/store/reblogs/index.ts
```diff
@@ -1,35 +1,41 @@
...
-      return state;
-  }
 };
 
 /* Actions */
 export const addReblog = (account: string, author: string, permlink: string) => (dispatch: Dispatch) => {
-  const data: Reblog[] = [...(ls.get("reblogs") || []), { account, author, permlink }];
-  ls.set("reblogs", data);
-  dispatch(reloadAct());
+    const data: Reblog[] = [...(ls.get("reblogs") || []), {account, author, permlink}];
```
feruzm, line 25: here cache we get from Feed page right? If post is opened directly with URL, this doesn't show user about reblog state, let's change that, there is already special call to get list of reblogs from api. 

`{"jsonrpc":"2.0", "method":"condenser_api.get_reblogged_by", "params":["good-karma","hivesigner-ui-simplified-import-and-login"], "id":1}`
This we can use when post is opened and not matter user state.

feruzm: we can save one request, if user coming from feeds by using this cache method still, but if coming from direct link, then we will have to do 1 more request to get reblogs list.

talhasch: > here cache we get from Feed page right? If post is opened directly with URL, this doesn't show user about reblog state, let's change that, there is already special call to get list of reblogs from api.
> 
> `{"jsonrpc":"2.0", "method":"condenser_api.get_reblogged_by", "params":["good-karma","hivesigner-ui-simplified-import-and-login"], "id":1}`
> This we can use when post is opened and not matter user state.

We store that cache in local storage. No matter page, that cache loaded from local storage to store on page load and updated once the user reblogs or deletes reblog.

feruzm: I tested above request, it seem to work quite fast so we might as well save user some storage if it works good. On feeds page we already have reblogged_by field and in post page, we can make 1 more request to fetch list. On mobile app, we will use this as well, instead of internal API we created for reblogs.

talhasch: I think you are referring to getting reblog count of a specific post. But this file (store/reblogs/index.ts) is responsible to manage reblogs that the user (or users) make. So we show the user if they reblogged a post (on post lists and post detail) or not.

feruzm: Right, with this call we can get count as well, as length of array. And know if user reblogged without caching. So I am thinking combining both solution seem good option. Or not cache reblogs at all. Which one you think is good?

talhasch: Is there an endpoint to get posts that the user reblogged?

feruzm: @talhasch I don't think it exist. Mostly post specific reblog info with above api call I shared `condenser_api.get_reblogged_by`

talhasch: @feruzm changed local storage caching to `condenser_api.get_blog_entries`

talhasch: Integrating `condenser_api.get_reblogged_by` to show rebloggers....

feruzm (approved):
lgtm

PR #423, opened 2021-04-01
minor style improvements

feruzm:


src/style/_markdown.scss
```diff
@@ -92,6 +92,9 @@
 
   img {
     max-width: 100%;
+    display: block;
```
talhasch, line 95: @feruzm why are we adding display and margins? i don't remember exactly but this attributes caused issues once.

feruzm: @talhasch Idea is to center images if they are not and margin auto works best with display block, you could try removing display block and see if it works as expected.

talhasch: this basically disallows users to add images aligned to left. removed styles. rest looks good.

feruzm: I did check pull-right and pull-left was working as expected and `<a>` style was about video embeds also looks better when they are centered. Centering images are really important if user doesn't use `<center>` or pull-right, pull-left then we should center it with styles.

talhasch (approved):

PR #427, opened 2021-04-05
markdown image styles

talhasch:


feruzm (approved):
ok, in another PR you will add centering right?!

PR #433, opened 2021-04-06
proxified images in gallery

talhasch:


feruzm:
@talhasch Gallery is only shown if `USE_PRIVATE = 1` right?!

talhasch:
yes, it is. but need to modify new mobile image dialog for that. will push another PR soon. you can merge this if it is ok.

feruzm (approved):
lgtm

PR #446, opened 2021-04-15
'Choose a community' by default

talhasch:


feruzm (approved):
lgtm

feruzm:
@talhasch Looks like there is bug, when select community from community list and then remove tags, default option falls back to My Blog instead of Choose a community. This works good if you add tags manually by typing and remove. But as soon as you add community from Modal list and then clear, it falls back to My Blog.

PR #447, opened 2021-04-15
account history filter update to fix desktop

talhasch:


feruzm (approved):
lgtm, let's test compiled version to make sure.

PR #452, opened 2021-04-19
Notifications

talhasch:


feruzm (approved):
lgtm

feruzm:
@talhasch btw, do we have use-private check on this? also what happens if there are no hivemind notifications but there are ecency notifications, does it give error?

talhasch:
we have usePrivate checks for notifications. 

we don't check if the user has notifications on hivemind. just sending an extra custom json once the user click "set all read" button.

PR #469, opened 2021-05-13
Get started banner issue fixed

kodcuserkan:
Fixed for logged-in users

### DIFF
--- a/src/common/components/intro/index.tsx
+++ b/src/common/components/intro/index.tsx
@@ -2,11 +2,15 @@ import React, { useEffect, useState } from 'react';
 
 import {Link} from "react-router-dom";
 
+import {connect} from "react-redux";
+
 import {Global} from '../../store/global/types';
 
 import {_t} from "../../i18n";
 
 import {closeSvg} from '../../img/svg';
+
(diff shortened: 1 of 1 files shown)


src/common/components/intro/index.tsx
```diff
@@ -19,7 +19,17 @@ interface Props {
 export default (props: Props) => {
     const hideIntro = () => props.hideIntro();
 
-    if (!props.global.intro) {
+    const [isLoggedIn, setIsLoggedIn] = useState(false)
+
+    useEffect(() => {
+        const verification = localStorage.getItem("ecency_active_user")
```
feruzm, line 25: I don't think you would need localStorage, you can use activeUser with props, no? Check `entry-index.tsx`

feruzm (changes requested):

src/common/components/intro/index.tsx
kodcuserkan: Ok will fix

kodcuserkan:
I did the changes used activeUser props to get current active user.

feruzm (approved):
lgtm

PR #471, opened 2021-05-14
Test fail fixed while hiing banner

kodcuserkan:


src/common/components/intro/index.tsx
```diff
@@ -18,23 +18,12 @@ const friendsWebp = require('../../img/welcome_community.webp');
 interface Props {
```
feruzm, line 18: cleanup unnecessary `imports` please.

feruzm (approved):
lgtm

PR #472, opened 2021-05-15
Bug/improve line height terms of use

kodcuserkan:
done for all static pages


feruzm:
On mobile still seem bit small, maybe increase a font size when on mobile and increase light heigh.

kodcuserkan:
H1 size increased and line height for texts increased
 Screen Shot 2021-05-16 at 01 30 53

feruzm (approved):
lgtm

PR #480, opened 2021-05-24
Improve/new landing page

feruzm:


.vscode/settings.json
```diff
@@ -0,0 +1,10 @@
+{
```
feruzm, line 1: I don't think it is good idea to add vscode settings into repository. Please remove this file/settings to keep it local.

feruzm:
lgtm, let's merge and do some test in production alpha

PR #498, opened 2021-06-02
Impovement/slider ux

mghazanfar:


feruzm:
@mghazanfar 

- Popup should be dismissible when clicked outside.
- Clicking on vote button shouldn't mark button blue
- Also check image, voting button marked blue and it has something on background making it look rectangle
 
image

mghazanfar:
> * st.
Status for the points above:

- Popup should be dismissible when clicked outside. (Done)

- Clicking on vote button shouldn't mark button blue, only after vote is successful. **(Done)**

- Also check image, voting button marked blue and it has something on background making it look rectangle probably because not using svg file or background of icon is white not transparent **(Done)**
- If post is already voted, slider shouldn't allow same percent voting, button within slider should be in disabled state perhaps, different percent is possible to cast on already voted post. **(In progress)**
- Also make sure it looks good for dark and light theme coloring. **(TBD)**

Latest code is pushed so you can test it meanwhile.

src/common/components/navbar/index.tsx
```diff
@@ -1,23 +1,23 @@
-import React, {Component} from "react";
+import React, { Component } from "react";
```
feruzm, line 1: I can't seem to find what exact line was changed due to linting changes, please revert to default linting if you can. @mghazanfar

mghazanfar: @feruzm I edited code for pencil button's placement and styling. I've disabled linting so wont be pushing these type of issues any more.

src/common/pages/submit.tsx
```diff
@@ -536,6 +536,7 @@ class SubmitPage extends BaseComponent<Props, State> {
                 addEntry(entry);
 
                 success(_t("submit.published"));
+                this.clear();
```
feruzm, line 539: Do same on Scheduling post...after success on api, clear.

src/common/components/profile-menu/index.tsx
```diff
@@ -46,15 +47,32 @@ export class ProfileMenu extends Component<Props> {
             label: ProfileFilter[section] ? _t(`profile.section-${section}`) : '',
             items: [
                 ...[ProfileFilter.blog, ProfileFilter.posts, ProfileFilter.comments, ProfileFilter.replies].map((x) => {
+                    let isActive = false;
+                        if(section===x && section ==='blog' && data.length!==0){
```
feruzm, line 51: please keep consistent spacing, `section===x` and `section ==='blog'` and `data.length!==0` all using different styling.

PR #505, opened 2021-06-07
Fix/no blogs

mghazanfar:
On visiting a profile, this PR covers:
- making "Posts" default option of profile menu if not blogs are present. (Only for first render)
- letting user move to blogs section manually
- showing "Nothing found" when user is not the loggedin person
- showing option to create a post if user visits their own profile
- a component with test for showing no data found


feruzm:
@mghazanfar Changing to Blog filter manually redirects back to Posts, we should probably let filter be if user sets it or selects that manually also when visits directly like `/@demodemo/blog`.

feruzm (approved):
lgtm

PR #522, opened 2021-06-11
Improve/feeds

mghazanfar:
For moving Friends and Communities to left, below are the tasks done:

**Basic Functionality**

- We now have 3: hot, trending and new for unauthorised user
- We have 4: friends, hot, trending and new for authorised user
- For unauthorized user, tabs navigate to global routes only
- For authorized user, tabs navigate to /my(ed) routes i.e. trending/my etc. We have a checkbox with label "global" that will open up global routes.

**Edge Cases**

- If a topic is selected, we show a cross with that.
- If a topic is selected, global gets checked.
- If a topic's cross is clicked, global is unchecked and my(ed) route is called.
- If a topic is selected and global is unchecked, selection of topic is discarded and user is redirected to my(ed) route
- If Friends is selected, we show no Global checkbox but state of the checkbox is maintained.
- If user is unauthenticated and select a topic then logs in, global checkbox is checked and that filtered route is shown for that user.

**New features**
- There should be a descriptive text about the filters applied.
- Text should remain same for global/local if filter is feed, otherwise text changes on with global checkbox's value
- Text is in a tooltip


feruzm (approved):
Lgtm, let’s do some tests

feruzm:
@mghazanfar tests failed in CI, please update tests

PR #524, opened 2021-06-14
Support language based validations

mghazanfar:
**This PR includes:**
- addition of validation messages in all the forms for required fields
- addition of utility function that can change HTML5 input's validation message
- using the above mentioned utility function in all the files having required input field


feruzm (approved):
lgtm, some inputs are missing but we can add them gradually

PR #525, opened 2021-06-15
Edit image like community with popup

mghazanfar:
This PR includes:
- A popup to edit the image
- Changing image from popup changes the image everywhere without reload
- Changing image from form changes the image everywhere without reload


src/common/components/profile-card/_index.scss
```diff
@@ -227,3 +227,27 @@
...
+  cursor: pointer;
+
+  &:hover {
+    box-shadow: $box-shadow;
+  }
+
+  svg {
+    height: 16px;
+  }
+}
```
feruzm, line 253: end file with new line

src/common/components/profile-card/index.tsx
```diff
@@ -36,11 +36,13 @@ import {
...
 } from "../../img/svg";
 
+import { EditPic } from '../community-card'
+
 interface Props {
     global: Global;
     history: History;
     activeUser: ActiveUser | null;
-    account: Account;
+    account: any;
```
feruzm, line 45: Why type `any`?

src/common/components/profile-edit/index.tsx
```diff
@@ -16,6 +16,7 @@ import {getAccount} from "../../api/hive";
 
 interface Props {
     activeUser: ActiveUser;
+    account: any;
```
feruzm, line 19: Also why `any`?

src/common/components/profile-edit/index.tsx
```diff
@@ -87,12 +96,12 @@ export default class ProfileEdit extends BaseComponent<Props, State> {
         }).then((account) => {
 
             // reload page to refresh profile image
-            if (activeUser.data.__loaded && activeUser.data.profile?.profile_image !== account.profile?.profile_image) {
-                setTimeout(() => {
-                    window.location.reload();
-                }, 500);
-                return;
-            }
+            // if (activeUser.data.__loaded && activeUser.data.profile?.profile_image !== account.profile?.profile_image) {
```
feruzm, line 99: reload not needed anymore right? we can remove this part instead of comment if that's the case

feruzm:
@mghazanfar also after update, I am getting error message. Top navigation User menu part not updating avatar, only after reload of page, maybe trigger re-render on navbar or user menu component for it to fetch new image?
 
image

src/common/components/profile-card/index.tsx
mghazanfar: Neither Account nor FullAccount | BaseAccount were working for getting profile info. I work on it more.

src/common/components/profile-edit/index.tsx
mghazanfar: Same issue as above.

mghazanfar:
> @mghazanfar also after update, I am getting error message. Top navigation User menu part not updating avatar, only after reload of page, maybe trigger re-render on navbar or user menu component for it to fetch new image?
> 
> image

On settings page, it changes as both the images. But when we upload image on profile page, data does not update immediately in the avatar inside navbar but it makes a call to getAccounts which changes picture in navbar after a few moments.

feruzm (approved):

PR #535, opened 2021-06-19
Update slider to have warning when user moving towards 0

mghazanfar:


feruzm:
@mghazanfar warning seem to be showing decrease from any value, not from voted value only.
if I voted 65% and open slider go to 80 and come back to 75, I see warning, it should probably only show when below 65% ?

also there is this bug with demo.com account. when coming very close to zero, there is NaN showing.
image

feruzm (approved):

PR #541, opened 2021-06-20
Fix wrong reference to posting key

4Ykw:
The posting key was being populated with the active key, instead of the former one.


feruzm (approved):
Lgtm, thank you for contribution

PR #548, opened 2021-06-26
Improvement/profile menu

mghazanfar:
This PR improves dropdown menu on:
- Profile page so it renders all options in the navbar for wide screens where it renders dropdown menu only when user has less space for options
- Communities page so it renders all options in the navbar for wide screens where it renders dropdown menu only when user has less space for options


feruzm (approved):
lgtm

PR #583, opened 2021-07-30
Fix desktop app auto scroll

mghazanfar:


src/common/components/entry-list-item/index.tsx
```diff
@@ -189,7 +189,7 @@ export default class EntryListItem extends Component<Props, State> {
         const cls = `entry-list-item ${promoted ? "promoted-item" : ""}`;
 
         return (
-            <div className={_c(cls)}>
+            <div className={_c(cls)} id={entry.permlink.replace(/[0-9]/g, '')}>
```
feruzm, line 192: for unique id, always use `${entry.author}-${entry.permlink}` author and permlink together.

src/desktop/app/components/navbar/index.tsx
```diff
@@ -201,19 +201,21 @@ export class NavControls extends Component<NavControlsProps> {
         // scroll to anchor element
         let href = (history as any).entries[1].pathname.split('/');
         if(href){
-            href = href[href.length - 1]
+            href = href[href.length - 2] + href[href.length - 1]
```
feruzm, line 204: This logic seem bit naive, we have comment links and using split by `/` wouldn't work always I think. For example, https://ecency.com/hive-167922/@claudio83/is-posh-token-underestimated#@acidyo/re-claudio83-qx493l would fail, of course from feed you don't get this link directly, but I want to make sure it works without unexpected issues.

PR #597, opened 2021-08-10
Bug: navbar stickyness gone

mghazanfar:


src/common/components/navbar/_index.scss
```diff
@@ -512,6 +512,11 @@ body.medium-zoom--opened {
   top: 0px;
   left: 0px;
   right: 0px;
+  z-index: 2000000000000000000;
```
feruzm, line 515: Wow, that's too big number... maximum z-index can be 2147483647. So how about we do maximum 100 and check all z-indexes when increasing priority, which ones should be on top, etc ? I don't think we have on single page more than 100 z-index, if there are we go to 200 maximum, etc. little bit of search and consideration with increasing z-index but at least we will have a system how to handle those.

mghazanfar: I researched on this and did not find index greater than 9000 but the issue was even 1 point less than this one was not working so some library's internally added index was causing this so had to make it this big.

feruzm: Interesting, I would prefer to use variables for example like this: it is defined in boostrap style files within our project, so we can reuse those.

image

I don't see any library having more than that, within our project. What I mean is we should avoid potential long term issues like this one, I see we already have unmanaged number of z-indexes and it is always hard to find issues when features are too many. Like in case of confirmation dialog getting behind form issue I discovered other day, many we won't discover if we keep using random numbers is what I am trying to say.

mghazanfar: Sounds like a plan. I'll refactor for sure.

feruzm (approved):

PR #599, opened 2021-08-11
Bug: Edit and Delete button on direct comment links

mghazanfar:


src/common/pages/entry.tsx
```diff
@@ -114,6 +122,10 @@ class EntryPage extends BaseComponent<Props, State> {
         const {location} = this.props;
         if (location.pathname !== prevProps.location.pathname) {
             this.ensureEntry();
+            if(location.pathname.includes('re-')){
```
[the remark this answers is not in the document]
feruzm: Because comment links are set by applications and yes we do set with `re-xxxxx` but other apps might set it differently breaking this condition.

src/common/pages/entry.tsx
```diff
@@ -119,10 +119,11 @@ class EntryPage extends BaseComponent<Props, State> {
...
     componentDidUpdate(prevProps: Readonly<Props>): void {
-        const {location} = this.props;
-        if (location.pathname !== prevProps.location.pathname) {
+        const { location } = this.props;
+        let entry = this.getEntry();
+        if (location.pathname !== prevProps.location.pathname && entry) {
             this.ensureEntry();
-            if(location.pathname.includes('re-')){
+            if(entry.parent_author){
                 let entry = this.getEntry();
```
feruzm, line 127: 2nd `this.getEntry` is not necessary, right?

feruzm: Ah it is local mapping get so if ensureEntry updates it, it will get updated object, reverting removed I just did.

feruzm:
related to this #620

PR #620, opened 2021-08-26
Add edit/delete buttons only

mghazanfar:
This PR:

- adds edit and delete buttons only

- If user navigates to a newly added comment, they see 404 as already happening on ecency.com, did not touch to fix this. Reloading the page works as ecency.com does i.e. no 404 after a while but respective comment page but with Edit/Delete buttons.

- Updating a comment shows immediate effects but when user navigates back to the parent post, it shows old value and after some time, it shows edited value. Same happens on ecency.com when user edits a comment, there are immediate effects on that page. Then if user goes to the comment page, old value is there, comes back, old value is there. It takes several moments to update there.

- Everything else is working as previously


feruzm (approved):
lgtm

PR #621, opened 2021-08-26
Hive Engine integration

tipyy:
Hello.

I started working on an Hive Engine integration for ecency.
For now it just displays tokens but I plan to add some actions in the future.

The code still needs some work at this point but if possible a little code review would be nice.

There is no translations, I don't really know what's the process is for that, should I just create the translation strings and english version?
I also have some failing snapshot tests but it doesn't seems to be related to my code changes.

There is only two API requests for now so I used simple axios requests but I've found a library `sscjs` that can be nice, is it something I can add or are you precautious when it comes to dependencies?

And finally, the UX/UI is maybe not that great, especially the mobile version. It works but can be improved.
I used a table to display the token infos, this is maybe not the best choice when it comes to mobile.
CSS and UX is clearly outside of my area of expertise, so a little help would ne nice.

### DIFF
--- a/src/common/components/wallet-hive-engine/_index.scss
+++ b/src/common/components/wallet-hive-engine/_index.scss
@@ -20,7 +20,7 @@
       height: 40px;
       width: 40px;
       border-radius: 50%;
-      background: #fff;
+      background: $pinkish-grey;
       background-position: 50% 50%;
       background-repeat: no-repeat;
       background-size: cover;
@@ -51,7 +51,7 @@
 
       @include themify(night) {
         background: #283241;
-        color: white;
+        color: $pinkish-grey;
       }
     }
   }


feruzm:
Looks good to me, I would suggest showing only token abbreviation and combine balances or just show one balance and rest information about token full name, description, staked balance could be shown inside `i` icon. Icon can be front of balance, because after balance, operations dropdown icon could be placed.

For operations (transfer, stake, unstake, etc.) have dropdown similar to how current Hive and Points wallet sections work. When we redesign later, we will improve it with addition like last market price and customization on user part. So we don't show all tokens right away but allow selective ones. All accessible/viewable in Add token modal. 

Screenshot 2021-08-28 at 06 57 30

tipyy:
A quick update on this pull request.

I modified the UI and it now looks like the unified wallet you mentioned earlier.
Here are some screenshots for day and night mode:

Screenshot 2021-08-31 at 16 35 11
Screenshot 2021-08-31 at 16 35 25

I really like it this way, especially on mobile mode.

I also added rewards and reward claiming:

Screenshot 2021-08-31 at 18 45 11

I don't know what is you politic when it comes to new features, do you want to have everything before shipping?
IMO it already brings some value and can be part of an upcoming release to collect feedbacks/bugs.

Next step for me will be to add stake/unstake actions.

As always, comments are welcome, so don't hesitate.

Thanks

feruzm:
@tipyy looking much better

Could we try removing/hiding ticker name after balances?
Balance could be aligned to right close to `i` icon and `i` icon could be moved front of balance.
What happens if you have more than 5 tokens pending reward, will it overcrowd the screen, how on mobile screens will it look?!

I think, we can merge and iterate these changes in steps for sure. We are planning to do release sometime next week, we will include this for sure, until then after we merge to development, we usually do some tests on alpha.ecency.com.

tipyy:
I implemented your suggestions and yeah it looks good :

Screenshot 2021-09-01 at 18 44 25

About rewards claiming you're right. I don't have that much tokens to claim but when I look at some other people account, it looks overwhelming, especially for mobile:

Screenshot 2021-09-01 at 18 45 03

What do you suggest (I like your UI suggestions :)), should Iwedisplay only a limited amount and add a claim all button or should we think of another design?

feruzm:
@tipyy There is function/api to claim all tokens at once right? 

We can try maybe like this:

`X tokens, Claim all` - this means multiple tokens (X number of tokens) has pending reward and you can claim them all in one click. 
`0.22 POB` - if there is only 1 token has pending reward then we can show amount and token ticker.

tipyy:
Yes there is an option to claim multiple tokens at once.
I'll try to implement that during the week end.

feruzm:
@tipyy Just wanted to follow up and see if there is anything we can help with?! We are planning on a release end of next week.

tipyy:
Sorry, I've been a little bit busy lately and didn't had a lot of time to work on it.
I added a method to claim multiple rewards at once, but I'm less confortable with the frontend part, if you want to help.

Also, it takes some time between the request being properly send and hive engine really taking into account the reward claim.
I'm trying to see if there is a way for the user to be aware of it.

PR #622, opened 2021-08-29
Fix prod build error for desktop app

mghazanfar:


feruzm (approved):
lgtm, lets do some test

PR #641, opened 2021-09-10
Add existing file

internetbuilder:
...


feruzm (changes requested):
electron is not necessary in package json, we have desktop app related packages in dev/desktop/package.json, other changes look ok but default port is 3000 which we got used to in our dev environment and we have some local host local storage data so port 3000 would be great change to if you could do.

package.json
```diff
@@ -80,6 +80,7 @@
     "@types/serialize-javascript": "^1.5.0",
     "@types/webpack-env": "^1.14.0",
     "babel-preset-razzle": "^4.0.5",
+    "electron": "^14.0.0",
```
feruzm, line 83: This is still not necessary, could you please remove this?

src/desktop/package.json
```diff
@@ -122,12 +122,13 @@
...
     "@babel/register": "^7.10.5",
     "@teamsupercell/typings-for-css-modules-loader": "^2.2.1",
+    "@webpack-cli/serve": "^1.5.2",
     "babel-plugin-dev-expression": "^0.2.2",
     "chalk": "^4.1.0",
     "concurrently": "^5.3.0",
     "cross-env": "^7.0.2",
     "detect-port": "^1.3.0",
-    "electron": "^8",
+    "electron": "^14.0.0",
```
feruzm, line 131: `src/desktop/package-lock.json` could be removed as we use `yarn` mostly, there is yarn lock file already. Also did you test with electron 14 all functionality? This is major version bump, need thorough testing.

PR #653, opened 2021-10-02
Add .circleci/config.yml

internetbuilder:


feruzm:
We already have CI and CD step with GitHub actions, circle CI is redundant.

PR #721, opened 2021-11-19
Fix bug

mghazanfar:
This PR:
- fixes the bug so correct filter data is fetched
- uses the loading state so no data can be filtered until previous fetched successfully


feruzm (approved):
lgtm

PR #724, opened 2021-11-21
Added distinction between vote change and removal

4Ykw:
Hi @feruzm, this is a better description of what happens behind the scenes when re-voting is 0% or changed the current vote from what was previsouly.

Tested on my local instance as working correctly.

Needs all the other languages to change `vote-warning` to `vote-remove` and add what it really means vote-warning in this case.

Cheers

### DIFF
--- a/src/common/i18n/locales/en-US.json
+++ b/src/common/i18n/locales/en-US.json
@@ -175,7 +175,8 @@
     "replies": "1 response. Click to respond.",
     "replies-n": "{{n}} responses. Click to respond.",
     "vote-error": "Previous value is not acceptable. Vote with a different value.",
-    "vote-warning": "Voting 0% unvotes the content.",
+    "vote-remove": "Voting 0% unvotes the content.",
+    "vote-warning": "Changing already voted content.",
     "no-replies": "No responses yet. Click to respond."
   },
   "nsfw": {
--- a/src/common/i18n/locales/en-US.json
+++ b/src/common/i18n/locales/en-US.json
@@ -176,7 +176,7 @@
     "replies-n": "{{n}} responses. Click to respond.",
     "vote-error": "Previous value is not acceptable. Vote with a different value.",
     "vote-remove": "Voting 0% unvotes the content.",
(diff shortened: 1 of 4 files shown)


src/common/i18n/locales/en-US.json
```diff
@@ -175,7 +175,8 @@
     "replies": "1 response. Click to respond.",
     "replies-n": "{{n}} responses. Click to respond.",
     "vote-error": "Previous value is not acceptable. Vote with a different value.",
-    "vote-warning": "Voting 0% unvotes the content.",
+    "vote-remove": "Voting 0% unvotes the content.",
+    "vote-warning": "Changing already voted content.",
```
feruzm, line 179: Thank you @4Ykw for your contribution.
- Could we change warning message to `Changing vote affects your curation reward.` ?
- Also could you add yourself in to contributors.json as contributor/developer?

4Ykw: The "vote-warning" is being displayed for both downvote and upvote. That message makes sense only for upvoting.
Shall I remove the warning message from when downvoting or create a distinct message for when downvoting?

Any default tests to run or all ? Last time I ran them on the development branch, there was a lot of things that were not passing, but it might be something I have on my testing environment. I will give it another go when possible and have a better understanding.

4Ykw: Uploaded new changes...

First time going through these tests workflow... but I guess all looks good with 2 changes that were not committed by me.

```
Snapshot Summary
 › 17 snapshots updated from 6 test suites.
 › 2 snapshots removed from 1 test suite.
   ↳ src/common/components/view-keys/index.spec.tsx
       • (1) Default render 1
       • (2) Dialog content 1

Test Suites: 97 passed, 97 total
Tests:       305 passed, 305 total
Snapshots:   2 removed, 17 updated, 290 passed, 307 total
Time:        7.269 s
Ran all test suites related to changed files.
```

Is this ok?

feruzm (approved):
lgtm, thank you!

PR #742, opened 2021-12-06
Support market page

mghazanfar:
Market page that enables internal exchange powered by blockchain.


src/common/api/hive.ts
```diff
@@ -12,6 +12,7 @@ import isCommunity from "../helper/is-community";
 
 import SERVERS from "../constants/servers.json";
 import { dataLimit } from './bridge';
+import moment from "moment";
```
feruzm, line 15: unnecessary?

package.json
```diff
@@ -46,18 +46,20 @@
     "path-to-regexp": "^6.1.0",
     "qs": "^6.9.4",
     "query-string": "^6.13.1",
-    "react": "^16.8.6",
+    "react": "^17.0.2",
```
feruzm: There are some breaking changes, could you check them? Event listeners likely need update. https://reactjs.org/blog/2020/10/20/react-v17.html#changes-to-event-delegation

mghazanfar: It was for trial. Reverted back. I thought that may be it has SSRSuspense available..

Issue #458, opened 2021-04-22
Price Feed Age Indication


Rishi556:
There's no indication of the age of the price feed on the witness page. With a lot of Hive's economy revolving around an up to date pricefeed from witnesses it would make sense to alert the users in some way of witnesses with older price feeds. Currently in the top 50 witnesses, there are 6 with price feeds which haven't been updated in a week.

Issue #457, opened 2021-04-22
No indication a witness is disabled


Rishi556:
Witnesses who are stale or disabled aren't shown differently, making it difficult to know for the user. An example would be netuoso, currently at rank 40, but has been disabled for over 3 months. There's no indication of that.

<img width="1105" alt="Screen Shot 2021-04-22 at 12 24 06 AM" src="https://user-images.githubusercontent.com/10539109/115660031-0e96f100-a301-11eb-840b-752b0dc0ef20.png">

PR #743, opened 2021-12-08
witnesses.tsx improved.

### LINKED ISSUE #456: Add Witness By Field
Not all witnesses operate a witness under their main account(or even similarly named accounts). Having a `by` field in the witness section can help users realize who the witness is by so they can better cast their votes. Example on peakd + condenser attached. Support for multiple users in the by field as a csv would be nice too(something condenser has but not peakd). The required data is stored in `witness_owner` field in the `profile` field of the account's metadata.

Screen Shot 2021-04-22 at 12 16 10 AM
Screen Shot 2021-04-22 at 12 16 33 AM

muhammadahmad57:
Fixes #456, #457, #458

### DIFF
--- a/src/common/pages/witnesses.tsx
+++ b/src/common/pages/witnesses.tsx
@@ -153,18 +153,17 @@ class WitnessesPage extends BaseComponent<PageProps, State> {
     }
 
     getWitness = async (witnessArray:WitnessTransformed[])=>{
-        const witnessUserNamesArray:string[] =
-            witnessArray.map((item: WitnessTransformed)=>{
+        const witnessUserNamesArray:string[] = witnessArray.map((item: WitnessTransformed)=> {
             return item.name
         })
         try {
             const accounts:FullAccount[] = await getAccounts(witnessUserNamesArray);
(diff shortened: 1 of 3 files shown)


src/common/pages/witnesses.tsx
```diff
@@ -144,6 +152,31 @@ class WitnessesPage extends BaseComponent<PageProps, State> {
...
+            return item.name
+        })
+        try {
+            const accounts:FullAccount[] = await getAccounts(witnessUserNamesArray);
+            const byWitnessState:WitnessTransformed[] = witnessArray.map((item: WitnessTransformed, index: number) => {
+                try {
+                    const parsedArray = JSON.parse(accounts[index].json_metadata ?accounts[index].json_metadata:'');
+                    return {
+                        ...item,
+                        witnessBy:parsedArray.profile.name ? parsedArray.profile.name: undefined
```
feruzm, line 167: Not `name` should be parsed, but `witness_owner` field should be parsed and not all witnesses have that field.

feruzm (approved):
lgtm

PR #745, opened 2021-12-13
Bug fix/title word break

adilrime17:
h1.entry-title word-break CSS changed to break-word.


feruzm (approved):
Lgtm, theme related changes will likely be reverted when we merge relevant PR

PR #746, opened 2021-12-14
Navbar search bugfixes on Electron

mghazanfar:


src/desktop/app/components/navbar/index.tsx
```diff
@@ -110,7 +110,16 @@ export class AddressBar extends Component<AddressBarProps, AddressBarState> {
         }
 
         if(location.search && location.search.length > 0){
-            address = location.pathname + location.search
+            address = location.pathname + location.search;
+            if(address[0]==="/"){
+                address = address.replace("/","")
+            }
+        }
+
```
feruzm, line 118: This would be better so we do `.split` once and reuse variable.
```
const sp = address.split('/');
if (sp[0] == sp[1]) {
address = sp;
...
```

feruzm: @mghazanfar 👆

feruzm (approved):
lgtm

PR #748, opened 2021-12-15
Feature/faq url search

mghazanfar:


feruzm (approved):
still one issue discovered, when language is different going to home with logo click will change dropdown language but interface stays with new temporary language until refresh. Minor but not urgent issue for now.

PR #755, opened 2021-12-23
Fix/auto complete

mghazanfar:


src/common/components/textarea-autocomplete/index.tsx
```diff
@@ -79,39 +81,32 @@ export default class TextareaAutocomplete extends BaseComponent<any, State> {
...
-								});*/
-							}
+								}, 300);
+							});
 						},
 						component: (props: any) => {
 							return (
 								<>
-									{props.entity.indexOf('/')===-1 && UserAvatar({ global: this.props.global, username: props.entity, size: "small" })}
+									{UserAvatar({ global: this.props.global, username: props.entity, size: "small" })}
```
feruzm, line 103: There is no avatar when selecting path, so there should be check if search is user or path. @mghazanfar

src/common/helper/posting.ts
```diff
@@ -42,7 +42,7 @@ export const createPermlink = (title: string, random: boolean = false): string =
 
 export const extractMetaData = (body: string): MetaData => {
     const urlReg = /(\b(https?|ftp):\/\/[A-Z0-9+&@#/%?=~_|!:,.;-]*[-A-Z0-9+&@#/%=~_|])/gim;
-    const userReg = /(^|\s)(@[a-z][-.a-z\d]+[a-z\d])/gim;
+    const userReg = /(^|[ \t]{2,})(@[a-z][-.a-z\d]+[a-z\d])(\/?[-a-z\d]*)/gm;
```
feruzm, line 45: This regex seem wrong and only catches few cases, I don’t think we need any change on regex here, because it is only responsible for adding mentions on post json-metadata.

feruzm (changes requested):

feruzm (approved):
lgtm

PR #758, opened 2021-12-25
Fix no show topics

mghazanfar:


feruzm (changes requested):
@mghazanfar It does fetch topics always unnecessarily. I think we should improve a logic bit further. For example for community feeds we don’t need topics because we don’t show them there. Maybe cache or keep topics in redux for an hour or until hare refresh so we don’t fetch them every time?!

mghazanfar:
> @mghazanfar It does fetch topics always unnecessarily. I think we should improve a logic bit further. For example for community feeds we don’t need topics because we don’t show them there. Maybe cache or keep topics in redux for an hour or until hare refresh so we don’t fetch them every time?!

Sure.

ghazanfar-BrilliantStar:
> @mghazanfar It does fetch topics always unnecessarily. I think we should improve a logic bit further. For example for community feeds we don’t need topics because we don’t show them there. Maybe cache or keep topics in redux for an hour or until hare refresh so we don’t fetch them every time?!

@feruzm I just went into details. Added my logic as well. But when I saw ``/ecency-vision/src/common/store/trending-tags/index.ts`` the logic is already there. If trending tags are already there then it does not refetch and recalls the API so I think we're good to go. It is already optimized.

feruzm (approved):

PR #791, opened 2022-01-11
Improve/no data

### LINKED ISSUE #789: Bug: search field on empty data
Don't show search field if there is no content to search from. If user posts are empty or comments are empty, no need to show search.
Same for community feeds, if community doesn't have any content yet, don't show search.

mghazanfar:
@feruzm This PR closes #789


feruzm (changes requested):
@mghazanfar Looks like message showing is irrelevant.

http://localhost:3000/created/hive-123321

```
Here you will see your posts, looks like no content exists.
No feed found.
```

http://localhost:3000/@abc/posts

```
Here you will see your posts, looks like no content exists.
No posts found.
```

I like actionable buttons on community page but for profile page, when other viewing that profile, they cannot create post for that user, so message is not relevant, misleading. For profile page we can show `Follow to know when they post` button where people/if logged in can follow.

feruzm (approved):
lgtm

PR #796, opened 2022-01-15
Feature/other user account wallet transfers

adilrime17:
I have completed this to the best of my knowledge of the app. Please review and test the implementation and let me know if I'm missing something.
Please test all the scenarios where Transfer component is used.
Best

### DIFF
--- a/src/common/components/wallet-ecency/index.tsx
+++ b/src/common/components/wallet-ecency/index.tsx
@@ -171,6 +171,7 @@ interface Props {
     signingKey: string;
     transactions: Transactions;
     fetchPoints: (username: string, type?: number) => void;
+    updateWalletValues: () => void;
     addAccount: (data: Account) => void;
     updateActiveUser: (data?: Account) => void;
     setSigningKey: (key: string) => void;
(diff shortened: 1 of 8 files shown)


src/common/components/wallet-ecency/index.tsx
```diff
@@ -281,20 +282,20 @@ export const WalletEcency = (props: Props) => {
...
+    //     history: history,
+    //     label: '',
+    //     items: [{
+    //         label: _t('points.transfer'),
+    //         onClick: toggleTransfer
+    //     }, {
+    //         label: _t('points.promote'),
+    //         onClick: togglePromote
+    //     }, {
+    //         label: _t('points.boost'),
```
mghazanfar, line 295: @adilrime17 could we not add the condition here that you added in the below?

mghazanfar (changes requested):

src/common/components/wallet-ecency/index.tsx
adilrime17: To make it consistent with the other component which is wallet-hive, I have added it below so developers can relate the code just by seeing it.
Let me know if you still want to change this.

mghazanfar (approved):
LGTM

feruzm (approved):
lgtm

Issue #792, opened 2022-01-11
Editor mention/path bug


feruzm:
List gets behind sometimes

![image](https://user-images.githubusercontent.com/3015990/148920809-fa9cfd87-c566-4453-8694-f690a3d7495a.png)

feruzm:
#797

PR #797, opened 2022-01-15
bug fixed by providing boundariesElement prop to plugin

adilrime17:
Studied the React Textarea Autocomplete plugin that we are using and it has a boundariesElement prop. Now the Autocomplete popup will stay in the editor even if the cursor is at the bottom or at the right mode side of the editor.


feruzm (approved):
lgtm, let's try it out in different scenarios. Closes #792

Issue #793, opened 2022-01-14
Images not rendering properly when viewed in full context


frittro:
The images in [this reply](https://ecency.com/hive-106316/@frittro/introducing-my-selfhosting-project-for#@frittro/re-ifarmgirl-2022114t131723388z) do not look right. This is after clicking the "View the full context" link. However, the same reply, when viewed [out of context](https://ecency.com/hive-106316/@frittro/re-ifarmgirl-2022114t131723388z), is rendering the images properly. The reply itself was created entirely in HTML, not in markdown, if that makes any difference. It also renders properly in [an alternate front-end](https://peakd.com/hive-106316/@frittro/introducing-my-selfhosting-project-for#@frittro/re-ifarmgirl-2022114t131723388z).

frittro:
Looks like this has nothing at all to do with the fact that I replied using HTML instead of markdown. I just made [this reply](https://www.ecency.com/blockchaingaming/@frittro/re-witherbloom-2022115t10012801z) on another thread, and the same thing happened to the icon there. I am re-posting the same icon link each time, to conserve RC on uploads.

feruzm:
@frittro Yes, that's like due to post and comments are being rendered in different style causing this inconsistency. We will fix it. When you open comment directly you are potentially using post renderer, but when you going to parent then you using comment renderer, so there is some difference bug that we need to fix.

feruzm:
Fix is already in production!

PR #798, opened 2022-01-15
Images rendering in comment section issue.

adilrime17:
Debugged and find that the CSS for the mini-markdown image needs to change. I have fixed the issue to the best of my understanding from the comments on https://github.com/ecency/ecency-vision/issues/793.
Let me know if something is missing.


feruzm (approved):
lgtm 👍

PR #810, opened 2022-01-21
Improve/witnesses

### LINKED ISSUE #775: Mobile friendly witnesses screen
Make /witnesses screen more mobile friendly.

mghazanfar:
This PR closes #775

### DIFF
--- a/src/common/components/witness-card/index.tsx
+++ b/src/common/components/witness-card/index.tsx
@@ -28,7 +28,7 @@ export const WitnessCard = ({
         </div>
         <div>
           <img
-            src={`https://images.ecency.com/${global.canUseWebp ? "webp" : ""}/u/${row.name}/avatar/medium`}
+            src={`https://images.ecency.com/${global.canUseWebp ? "webp/" : ""}u/${row.name}/avatar/medium`}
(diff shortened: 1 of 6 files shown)


src/common/components/witness-card/index.tsx
```diff
@@ -0,0 +1,79 @@
...
+            {row.witnessBy && (
+              <div className={"notranslate"}>
+                <small>by {row.witnessBy}</small>
+              </div>
+            )}
+          </h3>
+        </div>
+        <div>
+          <img
+            src={`https://images.ecency.com/webp/u/${row.name}/avatar/medium`}
```
feruzm, line 30: This should check if webp is supported or not, please use global.canusewebp check

src/style/_components.scss
```diff
@@ -15,6 +15,7 @@
 @import "../common/components/edit-history";
 @import "../common/components/dropdown";
 @import "../common/components/search-box";
+@import "../common/components/witness-card/";
```
feruzm, line 18: New component should have simple test always, please add.

feruzm (changes requested):

PR #818, opened 2022-01-27
Fix navbar logo href and blank page if user not logged in and go to user feed

### LINKED ISSUE #817: Bug: If a user is not login and goes directly to /@username/feed then blank page shows

adilrime17:
I try my best to test all the possible scenarios. Please review.
This PR Closes #817 #816


feruzm (approved):
lgtm

PR #822, opened 2022-01-31
Fix: profile preview

adilrime17:
Profile preview popover will only be shown if the user keeps the mouse on the username for a while otherwise it will not open.


feruzm (approved):
Lgtm

PR #823, opened 2022-02-01
Fix: 767 Increase emoji selection box

### LINKED ISSUE #767: Increase emoji selection box
Right now it displays 4 across and 5 down. So maybe 8 across and 5 down would be better.

adilrime17:
Emoji selection box increased. Earlier it was 4x4. Now it's 8x5.
This PR Closes #767


feruzm (approved):
lgtm

PR #828, opened 2022-02-01
Fix: logohref if on other user posts

adilrime17:
Fixed. Please review


feruzm (approved):
lgtm, using `isCommunity` or creating standard utility that we can reuse in future would be better though.

PR #832, opened 2022-02-02
Improve: isCommunity check is implemented to check community page

adilrime17:
Replace earlier check with isCommunity function that we already have.


feruzm (approved):
Lgtm

PR #845, opened 2022-02-04
Feature/pinning post

### LINKED ISSUE #774: Pinning post into blog
2 parts:

1. Operation is simply updating account `posting_json_metadata` with additional field `"pinned": "permlink"` example, permlink is post permlink. This operation should be added into 3 dot menu when you are viewing your own post.
2. Ability to show that pinned post if exist on account's `posting_json_metadata` as `pinned` field on top of all blog entries.

adilrime17:
Please review and test possible scenarios and let me know if anything is missing.
I have tested the feature to the best of my understanding.
This PR Closes #774


feruzm (approved):
lgtm

src/common/components/entry-menu/index.tsx
```diff
@@ -300,21 +378,78 @@ export class EntryMenu extends BaseComponent<Props, State> {
             ];
         }
 
-        if (this.canPinOrMute()) {
+        if (this.canPinBothOptions()) {
```
noumantahir, line 381: @feruzm can you textually elaborate this behaviour or is it possible to test it using demo.com, any community I can post to and this test this as moderator etc...

What I understand is, this is a special case when post is in community and author is also moderator trying to pin post.

feruzm: @noumantahir Yes, let's say @demo.com is moderator of Ecency community and makes post in that community. So he can potentially pin that post to Ecency community as well as he can pin to his own blog. He can pin any content in Ecency community published by others as well. Only his own can be pinned into his profile/blog. Web code logic is implemented in a bit complicated way for some reason, perhaps due to lack of understanding all scenarios.

Basically you can pin any of you own post into your blog. You can post any content in community page if you are moderator of that community. So unpinning also comes with those conditions.

Issue #873, opened 2022-04-09
Bug: Scroll position


feruzm:
Feed pages, in community page or profile or main page should keep scroll position. If you open post from couple screen down after scrolling and come back from post, scroll position should be kept same position.

PR #874, opened 2022-04-11
Allowed to persist scroll position

dkildar:
Added ability to persist scroll position in pages after going back from history.


feruzm (approved):
lgtm, if there is github issue, please mention them next time 😉  
Fixes #873

PR #895, opened 2022-05-07
Fixed the decks storage

dkildar:


feruzm (approved):
lgtm, solving no user use cases

PR #913, opened 2022-05-28
fix: fixed issue#912-enable description editing on content editing

talentdev1680:


src/common/pages/submit.tsx
```diff
@@ -260,10 +259,11 @@ class SubmitPage extends BaseComponent<Props, State> {
             }
 
             const {title, body} = entry;
+            let description = entry.json_metadata?.description || '';
```
feruzm, line 262: Here we can generate new description if empty, check body summary function.

feruzm (changes requested):

feruzm (approved):
lgtm

PR #915, opened 2022-05-30
fix: scheduling-noshow-onedit/description-add-jsonmetadata

talentdev1680:


feruzm (approved):
Lgtm, let’s do some tests

PR #919, opened 2022-06-07
Fixed editor page bottom actions

dkildar:


feruzm (approved):
lgtm

feruzm (changes requested):
editing area is too small on certain devices/tablets

PR #932, opened 2022-06-20
Profile page entry list and pinned entry issue

dkildar:
1. Profile page converted to functional;
2. Render function hard calculating moved to hooks and state.


feruzm (approved):
lgtm

PR #962, opened 2022-07-21
Am/notification tweaks

ArmishMunir:
notification hover fix


src/common/components/notifications/index.tsx
```diff
@@ -157,7 +157,7 @@ export class NotificationListItem extends Component<{
                                     entry: {category: 'category', author: notification.author, permlink: notification.permlink},
                                     afterClick: this.afterClick,
                                     children: <div className="markdown-view mini-markdown reply-body">
-                                        {postBodySummary(notification.body, 100)}
+                                        {postBodySummary(notification.body, 1000)}
```
feruzm, line 160: @ArmishMunir  What's reason behind this change?  I don't think we need that long details/summary

ArmishMunir: i've reverted it. it was just by mistake.

feruzm (approved):
lgtm

PR #1002, opened 2022-08-16
filter feed reblogs

Adesojisouljay:


src/common/pages/entry-index.tsx
```diff
@@ -24,25 +24,35 @@ interface Props extends PageProps {
...
 
 class EntryIndexPage extends Component<Props, State> {
     state: State = {
         entryList: [],
         promoted: [],
+        noReblog: false,
     };
 
     componentDidMount() {
+        console.log('componentDidMount');
```
feruzm, line 38: No need this log

src/common/components/entry-index-menu-dropdown/index.tsx
```diff
@@ -100,12 +104,22 @@ export const EntryIndexMenuDropdown = (props: Props) => {
...
+        ]
+  }
+
   const dropDownConfig = {
     history: null,
     label: (
       <div className='tagDropDown'>
         <span className='pl-2' />
-        {tag === "" ? _t('entry-filter.filter-global') : tag === 'my' ? _t('entry-filter.filter-community') : tag === 'today' ? _t('entry-filter.filter-today') : tag === 'week' ? _t('entry-filter.filter-week') : tag === 'month' ? _t('entry-filter.filter-month'): tag === 'year' ? _t('entry-filter.filter-year') : tag === 'all' ? _t('entry-filter.filter-alltime') : tag}
+        {tag === "" ? _t('entry-filter.filter-global') : tag === 'my' ? _t('entry-filter.filter-community') : tag === 'today' ? _t('entry-filter.filter-today') : tag === 'week' ? _t('entry-filter.filter-week') : tag === 'month' ? _t('entry-filter.filter-month'): tag === 'year' ? _t('entry-filter.filter-year') : tag === 'all' ? _t('entry-filter.filter-alltime') : tag === `@${activeUser.username}` ? (props.noReblog === true ? _t('entry-filter.filter-no-reblog') : _t('entry-filter.filter-with-reblog')) : tag}
```
feruzm, line 122: There is some logical issue here: For example, if I am logged in with @demo but want to see other user's feed (@ecency's feed) /@ecency/feed then I get in options `@ecency` and `Exclude reblogs`. Also looks like when logout we are not showing feed of any user, this is bug, maybe we can fix that here as well. Find out why it is not rendering/recognizing.

feruzm (changes requested):

Adesojisouljay:
This is noted, I will check and fix bug.

Thanks.

Adesojisouljay:
Issue has been resolved on a new branch.

PR #1021, opened 2022-09-13
Fix dropdown so it hides on selection on mobile and web both

mghazanfar:


feruzm:
@mghazanfar Thank you for your contribution, good to see you around. We have one more fix pending related to this, @Adesojisouljay will add that to this same branch

feruzm (approved):
lgtm, we will add improvement in separate PR

PR #1034, opened 2022-09-23
* allow to change the hive signer user to something else

shawnpringle:
make hive signer app user a variable in a configuration file


feruzm:
@shawnpringle good idea, could we make it environment variable and call it CLIENT_ID instead of hivesignerApp? That way even server side knows it and passes it in globalProps, check how USE_PRIVATE is set up, similar approach would be great and unified without client config files

mghazanfar (approved):
LGTM

mghazanfar:
> @shawnpringle good idea, could we make it environment variable and call it CLIENT_ID instead of hivesignerApp? That way even server side knows it and passes it in globalProps, check how USE_PRIVATE is set up, similar approach would be great and unified without client config files

i approved the PR but this change needs to be incorporated.

feruzm:
> > @shawnpringle good idea, could we make it environment variable and call it CLIENT_ID instead of hivesignerApp? That way even server side knows it and passes it in globalProps, check how USE_PRIVATE is set up, similar approach would be great and unified without client config files
> 
> i approved the PR but this change needs to be incorporated.

Great... It is really needed feature. For clean implementation, I would still prefer ENV route instead of creating another config file though. @shawnpringle Let me know if you could improve it otherwise I can arrange it in coming week.

feruzm:
Closing in favour of #1088

PR #1073, opened 2022-10-06
fixed defalt value and lg screen

Adesojisouljay:


src/common/pages/entry.tsx
```diff
@@ -159,7 +159,12 @@ class EntryPage extends BaseComponent<Props, State> {
     let replyDraft = ss.get(`reply_draft_${entry?.author}_${entry?.permlink}`);
     replyDraft = (replyDraft && replyDraft.trim()) || "";
 
-    this.setState({ isMounted: true, selection: replyDraft });
+    const wordsWithoutSpace: any = entry?.body.trim()?.split(/\s+/);
```
feruzm, line 162: this will crash the app when entry.body is undefined, that's why it was moved before, instead we have to make sure it renders only after it has entry.body perhaps...

feruzm (changes requested):

src/common/pages/entry.tsx
Adesojisouljay: Noted,  I'll move to ensure and rewrite logic

feruzm (changes requested):
let's find solution without use of localstorage, use state or redux to keep values...

Adesojisouljay:
This is noted, I am working on a solution.
Thanks 🙏

Adesojisouljay:
### What does this PR ?

On this PR we have adjusted the behavior of word count on different screen sizes.
Presently word count shows an initial zero value for few seconds before updating, also when entry is updated, word count is set to zero. This also has been solved, now word count is immediately updated in every scenario.

### Screenshot/video
[View on loom](https://www.loom.com/share/0c410ba21eda4aca94cc66179841709a)

feruzm (approved):
lgtm 👍

Adesojisouljay:
🙏

PR #1081, opened 2022-10-23
Feat/notifications drawer

Adesojisouljay:


feruzm (changes requested):
@Adesojisouljay please add this style changes to component style instead, because bootstrap style is library style which we should keep intact and add addition style changes to component level.

Adesojisouljay:
Noted, I'll do that
Thanks.

feruzm (approved):
lgtm

PR #1088, opened 2022-10-31
Development make hive signer account a configured variable

shawnpringle:
It works in my tests.  One caveat is the need to set base to the URL of the site in src/common/constants/defaults.json when setting a new HIVESIGNER_ID (CLIENT_ID) and HIVESIGNER_SECRET and also set USE_PRIVATE=0.  I've renamed the env CLIENT_ID to HIVESIGNER_ID, as it is related to HiveSigner.


.gitignore
```diff
@@ -18,3 +18,4 @@ src/config.ts
 .vscode
 
 dev.sh
+src/client_config.ts
```
feruzm, line 21: This is unnecessary, right?

src/common/components/login/index.tsx
```diff
@@ -383,13 +386,9 @@ export class Login extends BaseComponent<LoginProps, State> {
   };
 
   login = async () => {
+    const { hsClientId } = this.props.global;
     const { username, key } = this.state;
 
-    if (username === "" || key === "") {
```
feruzm, line 388: why these lines are being deleted?

src/desktop/app/components/navbar/index.tsx
```diff
@@ -519,6 +519,106 @@ export class NavBar extends Component<Props, State> {
               </ToolTip>
             </div>
 
+<<<<<<< HEAD
```
feruzm, line 522: merging lines should be fixed here

src/desktop/app/components/navbar/index.tsx
```diff
@@ -531,6 +631,7 @@ export class NavBar extends Component<Props, State> {
       </>
     );
   }
+>>>>>>> dd3c11b17991febd0e658a2e5ed97f98d42a88ce
```
feruzm, line 634: also merging lines/conflicts should be resolved

src/server/index.ts
```diff
@@ -89,4 +91,11 @@ server
   // For all others paths
   .get("*", fallbackHandler);
 
+if (config.hsClientId === "ecency.app" && defaults.base === "https://ecency.com" && config.hsClientSecret.length === 0) {
```
feruzm, line 94: what's the purpose of these lines, I assume it was for testing purpose?!

src/server/state.ts
```diff
@@ -41,6 +41,7 @@ export const makePreloadedState = async (req: express.Request): Promise<AppState
       ["iOS", "AndroidOS"].includes(getOperatingSystem(req.headers["user-agent"]))
     ),
     usePrivate: Boolean(parseInt(config.usePrivate, 10))
+    hsClientId: config.hsClientId
```
feruzm, line 44: line above should end with `,`

feruzm (changes requested):
overall looks better than previous #1034

src/server/index.ts
shawnpringle: The lines are to prevent invalid configurations from frustrating running and malfunctioning.  It is better to fail and close.  In the first branch, this is the default setting.  Once this case is taken into account any of these four settings will be an invalid configuration.

.gitignore
shawnpringle: That is something I overlooked.

src/common/components/login/index.tsx
shawnpringle: That was an accident.

src/server/index.ts
feruzm: Hmm, I think, it doesn't necessarily serve the purpose. Perhaps trying like, warning message respond is better?! Check here: `src/server/util.tsx`. @shawnpringle

shawnpringle:
I am confident that these lines of code would save anyone who uses this
feature headaches, but if it is a deal breaker, I'll remove it.

feruzm:
@shawnpringle one last thing, in readme could you replace `Hive Signer` with `Hivesigner`? Thanks

feruzm (approved):
lgtm 👍

PR #1097, opened 2022-11-07
hive engine chart

Adesojisouljay:


feruzm:
@Adesojisouljay What token chart is being displayed? Idea with this feature was slightly different.

1. No need new chart library, we already have highcharts integrated, please utilize that.
2. Idea was to have chart about each token just like Ecency mobile wallet page. Here I tried to make quick sketch: 
image
Use space between token label and balance and display chart/price info as background and on mobile browsers, we don't need to show charts.
3. For tokens that doesn't have market history or chart data, then show straight line basically. (non changing price)

Adesojisouljay:
I haven't pushed some commits, that's was a sample in first commits.

I will check previous library.

package.json
```diff
@@ -33,6 +33,7 @@
     "@webscopeio/react-textarea-autocomplete": "^4.8.1",
     "axios": "^0.21.2",
     "bs58": "^4.0.1",
+    "chart.js": "^3.9.1",
```
feruzm, line 36: chart.js is still being added, couldn't we use highcharts for this? @Adesojisouljay

src/common/components/hive-engine-chart/index.tsx
```diff
@@ -0,0 +1,230 @@
...
+
+const [close, setClose]: any = useState([])
+const [timeStamp, setTimeStamp]: any = useState([]);
+const [tokenSymbol, setTokenSymbol] = useState("SWAP.HIVE")
+const [prices, setPrices]: any = useState([])
+
+useEffect(() => {
+  const getMarketData = async (symbol: any) => {
+      const { data: history } = await axios.get(
+          `https://info-api.tribaldex.com/market/ohlcv`,
```
feruzm, line 20: Can we set all hive-engine servers inside servers.json or create he-servers.json inside constants folder? So we don't hard code all separately but from single place when issue occurs we can replace from single place, not from all random components.

feruzm (changes requested):

feruzm (approved):
lgtm, HIVE_ENGINE_RPC_URL and chartApi should be under one file perhaps in separate PR.

Adesojisouljay:
Yes, will move that shortly

PR #1109, opened 2022-11-17
Fixed swap form

dkildar:
What does this PR?
Added dynamic font size for amount control in swap form
Fixed loading state for sign buttons

Steps to reproduce
1. Open swap market form and try to resize window -> enter pretty long amount -> amount value longer than input element(Fixed by adding dynamic font size based on viewport for device less than `md` breakpoint);
2. Try to sign with key or HS in the same frame


feruzm (approved):
lgtm 👍

Issue #1053, opened 2022-09-27
Referral url param keeping


feruzm:
Respect or keep `?ref=username` and `?referral=username` in link in all path/location, when people come from other sources online, if they click signup, this field automatically should be detected.

feruzm:
@MTSaeed there is 3 dots options in content page or content list. `...` when clicked there is sharing link or copy link features. We also have profile referrals page where you can copy your referral link after you login.

PR #1111, opened 2022-11-18
Keeping referral name is URL, later used or signup

MTSaeed:
**What does this PR?**
When user try to share the post , it will append the username of active user in URL as optional parameter.
When any person open the link, it will extract the value of referral from the URL and store it in local storage.
When they click on sign up, the referral field is automatically detected by getting the value from the local stroage.
After signup, the referral name should be removed from the local storage.

**Steps to reproduce**

1. Copy the link of any post and open it in Incognito window. The URL of one post is pasted as an example:
http://localhost:3000/writing/@cheah/the-lies-of-diary-of-a-bomoh?referral=demo.com
2. Do some actions whatever you want like open another post, apply filter and other what you want.
3. When you click on signup , the referral field is automatically detected and filled by the username of the person who share the post to you. 

**Issue number** 

fixes #[1053](https://github.com/ecency/ecency-vision/issues/1053)

**Screenshots/Video**

https://user-images.githubusercontent.com/106739598/202697874-186e2f23-29c9-4f60-803c-a059a5de7b84.mp4

### DIFF
--- a/src/common/pages/sign-up.tsx
+++ b/src/common/pages/sign-up.tsx
@@ -10,6 +10,8 @@ import { PageProps, pageMapDispatchToProps, pageMapStateToProps } from "./common
 
 import ReCAPTCHA from "react-google-recaptcha";
 
+import * as ls from "../../common/util/local-storage";
+
 import Meta from "../components/meta";
 import Theme from "../components/theme/index";
 import NavBar from "../components/navbar/index";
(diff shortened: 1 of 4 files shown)


src/common/pages/sign-up.tsx
```diff
@@ -49,12 +48,19 @@ class SignUpPage extends Component<PageProps, State> {
...
 
   componentDidMount() {
-    const { location } = this.props;
+    const { location, history } = this.props;
     const qs = queryString.parse(location.search);
     if (qs.referral) {
       const referral = qs.referral as string;
       this.setState({ referral, lockReferral: true });
     }
+    if (ls.get("referral")) {
```
feruzm, line 57: if url param has referral we should make that priority than ls.

feruzm (changes requested):

src/common/pages/sign-up.tsx
MTSaeed: done @feruzm kindly check

feruzm (approved):
lgtm

PR #1114, opened 2022-11-20
Swap form fixed

dkildar:
**What does this PR?**
Fixed swap form signing by HS

**Steps to reproduce**
Open market swap form and try to sign with HS


feruzm (approved):
Lgtm

PR #1117, opened 2022-11-21
Enable login button if it is electron

MTSaeed:
Keep the login button enabled if is it Electron app. Do not disable in relation with reCaptcha


feruzm:
It is already resolved with state initial value setting on isVerified

PR #1125, opened 2022-11-25
Fixed swap form to asset and dropdown styles

dkildar:
**What does this PR?**
Fixed swap form to asset and dropdown styles

**Steps to reproduce**
Open market swap form and try to change from asset


feruzm (approved):
lgtm

PR #1128, opened 2022-11-26
Advanced trading/market mode

dkildar:


src/common/components/market-advanced-mode/stake-widget-header-options.tsx
```diff
@@ -0,0 +1,73 @@
+import React, { ChangeEvent, useState } from "react";
+import { _t } from "../../i18n";
+import { Button } from "react-bootstrap";
+import { allStakeSvg, buyStakeSvg, sellStakeSvg } from "../../img/svg";
+
+export enum StakeWidgetViewType {
+  All = "all",
+  Sell = "sell",
+  Buy = "Buy"
```
feruzm, line 9: buy capitalized, on purpose?

tsconfig.json
```diff
@@ -42,12 +35,8 @@
     "jest",
     "razzle.config.js",
     "healthCheck.js",
-    "src/desktop/**/*.js"
+    "src/desktop/**/*.js",
+    "node_modules/react-use"
```
feruzm, line 39: Why do we need to add this here?
I suspect warning in console, coming from here...
Screenshot 2023-01-14 at 07 30 23

dkildar: Not from here. Warning because we have too old typescript version

dkildar: I tried to fix it but it didn't help. Will return as it was

src/common/components/market-advanced-mode/stake-widget-header-options.tsx
dkildar: Fixed

feruzm (approved):
🎉🛫

PR #1129, opened 2022-11-26
Add Brazilian Real (BRL) to currencies

igormuba:
I tested locally and it works. Brazilian users will be glad to see our currency added to Ecency.


feruzm (approved):
lgtm

PR #1137, opened 2022-12-02
Postingflags

shawnpringle:
I'll be adding these features:
show whether a post was self-voted, how its reward will be split, and when to display what the reward limit is.

Sometimes it is interesting to some to see whether the user has self-voted a post. Whether they are powering up 100% or whether they have limited their payout to some small value.

These are set in the site wide defaults file:
"showSelfVote": true, "showRewardSplit": true, "lowRewardThreshold": 1000000000
For example, if you set lowRewardThreshold to 1000, it wont show unless the post is restricted to 1000 HBD or to some other lower value.

See also: https://www.proofofbrain.blog/hive-150329/@leprechaun/show-self-vote-maximum-payout-hive-150329


src/common/components/comment/__snapshots__/index.spec.tsx.snap
```diff
@@ -270,6 +270,11 @@ exports[`(1) Default render 1`] = `
   >
     <div
       className="rta  "
+      style={
+        Object {
+          "height": "80px",
+        }
+      }
```
dkildar, line 277: What the `Object`? Just `{height: "80px"}`

src/common/components/comment/__snapshots__/index.spec.tsx.snap
```diff
@@ -602,6 +610,11 @@ exports[`(2) Cancellable, in progress 1`] = `
   >
     <div
       className="rta  "
+      style={
+        Object {
+          "height": "80px",
+        }
```
dkildar, line 616: Same as above

src/common/components/entry-list-item/index.tsx
```diff
@@ -286,6 +287,13 @@ export default class EntryListItem extends Component<Props, State> {
 
     const cls = `entry-list-item ${promoted ? "promoted-item" : ""} ${global.filter}`;
 
+    const self_vote_entry = entry.active_votes.find((x) => x.voter == entry.author);
+    const self_vote = self_vote_entry ? self_vote_entry.rshares : false;
+    const hp_portion = 100 * (1 - entry.percent_hbd / 20000);
```
dkildar, line 292: We use `camelCase` not `snake_case` in our project

src/common/components/entry-list-item/index.tsx
```diff
@@ -286,6 +287,13 @@ export default class EntryListItem extends Component<Props, State> {
 
     const cls = `entry-list-item ${promoted ? "promoted-item" : ""} ${global.filter}`;
 
+    const self_vote_entry = entry.active_votes.find((x) => x.voter == entry.author);
+    const self_vote = self_vote_entry ? self_vote_entry.rshares : false;
```
dkildar, line 291: `const selfVote = self_vote_entry?.rshares ?? false;`

src/common/components/entry-list-item/index.tsx
```diff
@@ -286,6 +287,13 @@ export default class EntryListItem extends Component<Props, State> {
...
+    const hp_portion = 100 * (1 - entry.percent_hbd / 20000);
+    const max_payout: number = parseFloat(entry.max_accepted_payout);
```
dkildar, line 293: No need extra `number`

src/common/components/entry-list-item/index.tsx
```diff
@@ -361,6 +369,24 @@ export default class EntryListItem extends Component<Props, State> {
...
+            {app && (
+              <>
+                &ensp;
+                <span itemProp="publisher" itemScope={true} itemType="http://schema.org/Person">
+                  <meta itemProp="name" content={entry.author} />
+                </span>
+                <div className="app" title={app}>
+                  <Tsx k="entry.via-app" args={{ app: appShort }}>
+                    <a href="/faq#source-label" />
+                  </Tsx>
```
dkildar, line 383: I guess `<a href="/faq#source-label">{_t('entry.via-app', { app: appShort })}</a>` much easier

src/common/components/entry-read-time/index.tsx
```diff
@@ -3,13 +3,25 @@ import { OverlayTrigger, Tooltip } from "react-bootstrap";
...
+  const { showSelfVote, showRewardSplit, lowRewardThreshold } = global;
 
   const [readTime, setReadTime] = useState(0);
   const [wordCount, setWordCount] = useState(0);
 
+  const self_vote_entry = entry.active_votes.find(
+    // @ts-ignore
+    (x) => x.voter == entry.author
+  );
+  const self_vote = self_vote_entry ? true : false;
```
dkildar, line 19: `const selfVote = !!selfVoteEntry`

src/common/components/preferences/index.tsx
```diff
@@ -98,6 +101,34 @@ export class Preferences extends BaseComponent<Props, State> {
     success(_t("preferences.updated"));
   };
 
+  showSelfVoteChanged = (e: React.ChangeEvent<typeof FormControl & HTMLInputElement>) => {
+    const { setShowSelfVote } = this.props;
+    const { value } = e.target;
+    console.log(value);
```
dkildar, line 107: Unnecessary log

src/common/components/preferences/index.tsx
```diff
@@ -98,6 +101,34 @@ export class Preferences extends BaseComponent<Props, State> {
...
+    success(_t("preferences.updated"));
+  };
+
+  lowRewardThresholdChanged = (e: React.ChangeEvent<typeof FormControl & HTMLInputElement>) => {
+    const { setLowRewardThreshold } = this.props;
+    const { value } = e.target;
+    try {
+      setLowRewardThreshold(JSON.parse(value));
+    } catch (e) {
+      console.log(value, "Couldn't get it done boss");
```
dkildar, line 127: We have a `error` function that will show an alert on the bottom right side of viewport. Please use it instead

dkildar: Error message not meaningful

src/common/components/preferences/index.tsx
```diff
@@ -114,6 +145,10 @@ export class Preferences extends BaseComponent<Props, State> {
     if (use_system_theme) {
       theme = "system";
     }
+    this.showSelfVoteChanged = this.showSelfVoteChanged.bind(this);
+    this.showRewardSplitChanged = this.showRewardSplitChanged.bind(this);
+    this.lowRewardThresholdChanged = this.lowRewardThresholdChanged.bind(this);
```
dkildar, line 150: Why you did this? Arrow functions already have a current context

feruzm:
@shawnpringle I would prefer bit discreet design to show these info. Can you check mobile view in posts, there we have `i` info icon that shows read time and word count, could we show these details inside that type of popover?! 
Idea of simplicity is to keep page from too much information being shown to new users. For many people these information has less value, imo. What do you think?

Issue #1048, opened 2022-09-27
Store post and comment vote percentages 


feruzm:
in localstorage separately.

MTSaeed:
Start slider from last voted percentage on comments and posts

PR #1138, opened 2022-12-02
Store post and comment vote percentages

MTSaeed:
**What does this PR?**

It will store the last vote percentage of post and comment in local storage.
If there is previously stored value in local storage , it will get the value from local storage and display the previous voted value.
If there is no value in local storage means user open the voting slider first time, it will display the default value in the slider.

**Steps to reproduce**

1. Open the voting slider of any post and vote to that post or just move the slider to change the value of slider. it will store the updated value of slider in local storage.
2. When you open the slider of any post next time, it will show and  set the previous stored value.
3. Same is the case with comment-voting-slider.But post-voting-slider value and comment-voting-slider value stored separately and  independent to each other.
4. Stored the value of slider in both **Up** and **Down** case.

**Issue number** 

fixes #[1048](https://github.com/ecency/ecency-vision/issues/1048)

**Screenshots/Video**

https://user-images.githubusercontent.com/106739598/205293384-87226905-d016-405f-94e4-9034421292a9.mp4

### DIFF
--- a/src/common/components/discussion/index.tsx
+++ b/src/common/components/discussion/index.tsx
@@ -418,7 +418,8 @@ export const Item = (props: ItemProps) => {
                     {EntryVoteBtn({
                       ...props,
                       entry,
-                      afterVote: afterVote
+                      afterVote: afterVote,
+                      isPostSlider: false
                     })}
                     {EntryPayout({
                       ...props,
--- a/src/common/components/entry-list-item/index.tsx
+++ b/src/common/components/entry-list-item/index.tsx
@@ -516,6 +516,7 @@ export default class EntryListItem extends Component<Props, State> {
           <div className="item-controls">
             {EntryVoteBtn({
               ...this.props,
+              isPostSlider: true,
               afterVote: this.afterVote
             })}
             {EntryPayout({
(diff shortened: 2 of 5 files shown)


feruzm:
@MTSaeed Do we still show voted value? If content already has been voted, slider should show that value not localstorage value.

MTSaeed:
@feruzm @dkildar The code has been updated please review. 

Currently we are saving the voted value in localstorage and showing that on slider next time if the user has voted on one post or comment. If the user just changes the value on slider and has not voted the value is not saved and the voted one will be displayed next time.

feruzm:
@MTSaeed I am talking about, content is voted by user with 40% and later another content was voted with 50%. If person opens first content vote again, do we show 40% or 50% ?

MTSaeed:
> @MTSaeed I am talking about, content is voted by user with 40% and later another content was voted with 50%. If person opens first content vote again, do we show 40% or 50% ?

It would say 40% as that post has been voted 40%. 

https://user-images.githubusercontent.com/106739598/206150706-e7d6510d-77cf-480e-8768-d615a5c50428.mp4

If a post is voted it would show the voted value instead of LS.

feruzm (approved):
lgtm, will require more testing on production alpha https://alpha.ecency.com

PR #1139, opened 2022-12-02
Fixed swap form price calculation

dkildar:
**What does this PR**
Fixed swap form price calculation

**Steps to reproduce**
Open swap form and try to swap. It should be finished very fast


feruzm (approved):
lgtm

Issue #1147, opened 2022-12-07
Bug: Editor Advanced section images are not showing


feruzm:
In editor advanced section, images are not showing to select thumbnail of the post. So post creation not allowing thumbnail selection, it was working before, perhaps some new changes, removed or broke it. Find out and improve logic.

igormuba:
Fix at
https://github.com/ecency/ecency-vision/pull/1148

feruzm:
@igormuba next time, in PR description or commits messages, add `Fixes #1147` message referencing this issue number. Once PR merged, it should automatically close this issue as well with it.

PR #1148, opened 2022-12-07
Fix for missing thumbnails on advanced editor

igormuba:
Fix for the issue
https://github.com/ecency/ecency-vision/issues/1147

Just need to comment back into the code the line that outputs the images from the post

### DIFF
--- a/src/common/pages/submit.tsx
+++ b/src/common/pages/submit.tsx
@@ -826,7 +826,7 @@ class SubmitPage extends BaseComponent<Props, State> {
 
   buildMetadata = () => {
     const { tags, body, description, selectedThumbnail, selectionTouched } = this.state;
-    const meta = extractMetaData(body);
+    const { ["thumbnails"]: thumbnails, ...meta } = extractMetaData(body);
     let localThumbnail = ls.get("draft_selected_image");
 
     if (meta.image) {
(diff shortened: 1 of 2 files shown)


igormuba:
Result of the change, allowing thumbnails to be seen:
image

feruzm (changes requested):
Right, uncommenting fixes the issue. Reason I commented that out however was due to addition of thumbnails into json_metadata of the post object to blockchain which is unnecessary. We can just re-arrange images array and not include extra thumbnails array into posting json. So it needs improvements...

src/common/pages/submit.tsx
```diff
@@ -826,7 +826,7 @@ class SubmitPage extends BaseComponent<Props, State> {
 
   buildMetadata = () => {
     const { tags, body, description, selectedThumbnail, selectionTouched } = this.state;
-    const meta = extractMetaData(body);
+    const { ["thumbnails"]: thumbnails, ...meta } = extractMetaData(body);
```
feruzm, line 829: @igormuba 
`const { thumbnails, ...meta} = extractMetaData(body)` should work as well, right?

igormuba: Yes, both have the same effect. Just to be sure I tried your way and it worked so I changed it, looks cleaner than my original solution, has the same effect

feruzm (approved):
lgtm

Issue #1149, opened 2022-12-07
Animation after vote is cast


feruzm:
Right now, we have blinking animation while vote is processed/submitted, try to improve that after vote cast with animation.
We increase heart number so animation could be checked to do on heart icon or on vote icon. Best to see different variations and see if there can be structure for that animation to be configurable in future.

Something like this: 
https://lottiefiles.com/88955-reddit-up-vote
https://lottiefiles.com/96937-like-heart
https://lottiefiles.com/88219-upvote-splash-animation
https://lottiefiles.com/9729-upvote

igormuba:
Hi, I am not sure if I understood the task.
I didn't understand the wording "We increase heart number ", what is that supposed to mean?

Other than that, in general, is it just to make it more clear that an animation is happening?

igormuba:
At 
https://animista.net/play/basic/rotate-scale
I found some easy to use animations

I just had to add to the file src/style/_animation.scss
```
@-webkit-keyframes rotate-scale-up {
  0% {
    -webkit-transform: scale(1) rotateZ(0);
    transform: scale(1) rotateZ(0);
  }
  50% {
    -webkit-transform: scale(2) rotateZ(180deg);
    transform: scale(2) rotateZ(180deg);
  }
  100% {
    -webkit-transform: scale(1) rotateZ(360deg);
    transform: scale(1) rotateZ(360deg);
  }
}
@keyframes rotate-scale-up {
  0% {
    -webkit-transform: scale(1) rotateZ(0);
    transform: scale(1) rotateZ(0);
  }
  50% {
    -webkit-transform: scale(2) rotateZ(180deg);
    transform: scale(2) rotateZ(180deg);
  }
  100% {
    -webkit-transform: scale(1) rotateZ(360deg);
    transform: scale(1) rotateZ(360deg);
  }
}
```

And then on src/style/_mixins.scss I on line 355 where it says `&.in-progress {` I add the new animation to play only once and leave the blinking animation playing for as long as it takes:
```
&.in-progress {
    animation: rotate-scale-up 0.65s linear both, anim-fadein-out 1s infinite;
    pointer-events: none;
  }
```

I like how subtly the original animation is but it is possible to make it more appearant, just need to decide what to do with it

feruzm:
Animations are something to be considerate about and we would like to add more of them in future. So to keep that in we need to go with most scalable solution that doesn't increase much of overall bundle size as well.

[...]

PR #1153, opened 2022-12-08
scss file for animations

igormuba:
Added a file specifically for animations, related to the issue
https://github.com/ecency/ecency-vision/issues/1149


dkildar:
We have the `_animation.scss`. What the reason to creating of `_animations.scss`?

igormuba:
> We have the `_animation.scss`. What the reason to creating of `_animations.scss`?

I was talking with @feruzm and we are thinking about adding animations that could or could not be used later. I am adding and testing a few animations (a few didn't work idk why) and will keep them in there in `_animations.scss`, separated from the `_animation.scss` that is in use

I am open to suggestions, thought, if someone has a different idea and wants to share.

feruzm:
@igormuba we can combine those into single _animations.scss and use that going forward. I forgot about animation.scss when I commented. Idea is to keep all animations into single place organized.

igormuba:
I updated and added some comments on how to use the new animations, left the originals but moved everything to _animations.scss

dkildar:
lgtm

feruzm (approved):
lgtm, let's try out in action and see if animations improves ux

PR #1169, opened 2022-12-16
Fixed market swap form price calculation

dkildar:
**What does this PR**
Fixed market swap form price calculation

**How to reproduce**
Open market swap form and see that price is not correct


feruzm (approved):
Lgtm

Issue #923, opened 2022-06-10
RAW content viewing, `?raw` also button on footer


feruzm:
show content in raw format, `?history` shows history of edits so raw format is basically complimenting that.

PR #1175, opened 2022-12-22
Post in raw format

MTSaeed:
**What does this PR?**

Show content of post and comments in raw format when user enter ?raw in post link.

**Steps to reproduce**

1. Open any post and append ?raw at the end of the link of that post.

**Issue number** 

fixes #[923](https://github.com/ecency/ecency-vision/issues/923)

**Screenshot / Video**

https://user-images.githubusercontent.com/106739598/209131616-5c5d58d4-c42d-4639-adc4-5b3b3eef1188.mp4


MTSaeed:
@feruzm One more thing i noticed and then try to correct is if user enter both two query parameters ?history and ?raw at the same time, then it would have to show the post in raw format and also the history of post.
what's your suggestion on this?.
You can see in the video given below:

https://user-images.githubusercontent.com/106739598/209136501-60089507-cbe3-4811-bef2-032cb3131631.mp4

feruzm:
@MTSaeed no need to support both, just make sure they are supported separately in clean way, if both exist in query just select last one as primary or use ?history as priority.

MTSaeed:
> @MTSaeed no need to support both, just make sure they are supported separately in clean way, if both exist in query just select last one as primary or use ?history as priority.

Updated.

feruzm:
@MTSaeed you forgot to add `<>` button on post page that will toggle `raw` just for main content.

src/common/img/svg.tsx
```diff
@@ -1731,6 +1731,7 @@ export const rawContentSvg = (
     xmlns="http://www.w3.org/2000/svg"
     width="23"
     height="23"
+    cursor={"pointer"}
```
feruzm, line 1752: @MTSaeed add cursor into style of component where this svg is used.

src/common/i18n/locales/en-US.json
```diff
@@ -465,7 +465,8 @@
     "community-in": "in",
     "post-word-count": "Words:",
     "post-read-time": "Reading:",
-    "post-read-minuites": "min"
+    "post-read-minuites": "min",
+    "raw": "Raw"
```
feruzm, line 469: @MTSaeed Text can be, `View RAW format`

MTSaeed:
Updated.

feruzm (approved):
lgtm

PR #1177, opened 2022-12-27
Fixed market swap form order books

dkildar:
**What does this PR**
Fixed market swap form price calculation

**How to reproduce**
Open market swap form and try to swap


feruzm (approved):
Lgtm

PR #1180, opened 2023-01-02
proposal votes pagination

### LINKED ISSUE #1161: Proposal votes pagination
Add pagination to proposal votes, as of right now, it doesn't seem to fetch all votes and paginate.

MTSaeed:
**What does this PR?**

Shows the HP of that proposal in the modal header.
Creates a search bar below the modal header to search the particular voter.
Shows the pagination and sorting option below the modal body.
when user reaches to the last page of pagination , it will fetches the next limited records until all the records fetched.
When user searches any voter , if that voter is not present in current list , it will fetch the next record and find that user until all the records fetched.

**Steps to reproduce**

1. Click on the "Votes" of any proposal , it will open the modal.
2. Search any record by typing name of voter in the search bar.If it is present in the current list , it will display that record 
    otherwise it will fetch the next records until all records been fetched.
3. User can jump into the last and first page by clicking on the last and first page icon.

**Issue number** 

Fixes #1161

**Screenshot / Video**

https://user-images.githubusercontent.com/106739598/210304914-d79b1361-48d9-434c-abe9-7aedfed992c0.mp4

### DIFF
--- a/src/common/components/proposal-votes/index.tsx
+++ b/src/common/components/proposal-votes/index.tsx
@@ -117,13 +117,13 @@ export class ProposalVotesDetail extends BaseComponent<Props, State> {
     const { voter, limit } = this.state;
     getProposalVotes(proposal.id, voter, limit)
       .then((votes) => {
-        getVotesCount(votes.length);
         this.setState({ voter: votes[votes.length - 1].voter, lastDataLength: votes.length });
         const usernames = [...new Set(votes.map((x) => x.voter))]; // duplicate records come in some way.
 
         return getAccounts(usernames);
       })
       .then((resp) => {
+        getVotesCount(resp.length);
         let voters: Voter[] = resp
           .map((account) => {
             const hp = (parseAsset(account.vesting_shares).amount * hivePerMVests) / 1e6;
@@ -294,6 +294,7 @@ export class ProposalVotes extends Component<ProposalVotesProps, ProposalVotesSt
   render() {
     const { proposal, onHide } = this.props;
     const { searchText, voteCount } = this.state;
+    const modelTitle = voteCount % 1000 ? voteCount + " " : voteCount + "+" + " ";
     return (
       <Modal
         onHide={onHide}
@@ -305,7 +306,7 @@ export class ProposalVotes extends Component<ProposalVotesProps, ProposalVotesSt
       >
         <Modal.Header closeButton={true} className="align-items-center px-0">
           <Modal.Title>
-            {voteCount + " " + _t("proposals.votes-dialog-title", { n: proposal.id })}
+            {modelTitle + _t("proposals.votes-dialog-title", { n: proposal.id })}
           </Modal.Title>
         </Modal.Header>
         <Form.Group className="w-100 mb-3">
--- a/src/common/components/proposal-votes/index.tsx
+++ b/src/common/components/proposal-votes/index.tsx
@@ -1,6 +1,6 @@
 import React, { Component } from "react";
 
-import { Modal } from "react-bootstrap";
+import { Modal, Form, FormControl } from "react-bootstrap";
 
 import numeral from "numeral";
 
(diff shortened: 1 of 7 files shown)


src/common/components/proposal-votes/index.tsx
```diff
@@ -29,42 +30,100 @@ interface Voter {
...
   onHide: () => void;
 }
 
 interface State {
+  page: number;
   loading: boolean;
   voters: Voter[];
+  originalVoters: Voter[];
+  sort: SortOption;
+  noOfPages: number;
```
dkildar, line 52: Just `pageCount`

src/common/components/proposal-votes/index.tsx
```diff
@@ -29,42 +30,100 @@ interface Voter {
...
+
+  search = () => {
+    const { originalVoters } = this.state;
+    if (this.props.searchText) {
+      this.setState(
+        {
+          voters: this.state.originalVoters.filter((item) =>
+            item.name.toLocaleLowerCase().includes(this.props.searchText.toLocaleLowerCase())
+          ),
+          page: 1
```
dkildar, line 93: Should it be `this.state.page` instead of 1?

src/common/components/proposal-votes/index.tsx
```diff
@@ -83,67 +142,151 @@ export class ProposalVotesDetail extends BaseComponent<Props, State> {
...
+    const start = (page - 1) * pageSize;
+    const end = start + pageSize;
+    const sliced = voters
+      .sort((a, b) => {
+        const keyA = a[sort]!;
+        const keyB = b[sort]!;
+
+        if (keyA > keyB) return -1;
+        if (keyA < keyB) return 1;
+        return 0;
```
dkildar, line 185: I guess you can simplify it: `voters.sort((a,b) => a[sort] - b[sort])`

src/common/components/proposal-votes/index.tsx
MTSaeed: No, it should be 1. Because we want to show the result of search on the 1st page.

src/common/components/proposal-votes/index.tsx
MTSaeed: If we use this method. There is a TS error as it could not recognize what would be the value of a[sort], because a is an object and could not determine what key would be there in sort. Therefore we could not use this method. 

However I have tried summing this up with simple logical operator such as voters.sort((a,b) => +(a[sort] > b[sort])). But we would miss the case where a[sort] == b[sort]. Therefore I think we should keep this as it is. Please suggest if you have any other solution in mind.

feruzm:
@MTSaeed I was expecting we will show number of votes on title, not how much HP voted the proposal because that info already visible before modal is opened, so it is not needed to on modal title, instead showing number of votes would be interesting to see. Could we change that?!

MTSaeed:
> @MTSaeed I was expecting we will show number of votes on title, not how much HP voted the proposal because that info already visible before modal is opened, so it is not needed to on modal title, instead showing number of votes would be interesting to see. Could we change that?!

proposal's-test


https://user-images.githubusercontent.com/106739598/212839422-4c694e21-c925-41ec-9a63-c495f5d548f9.mp4

MTSaeed:
@feruzm Please check

feruzm:
@MTSaeed lets add `+` after vote count, like 1000+, 2000+, 3000+ but when 3198 then there is no `+` because it is end of list, makes sense? when there is more data available show or add `+` at the end of numbers on title when all fetched, no need +.

MTSaeed:
> @MTSaeed lets add `+` after vote count, like 1000+, 2000+, 3000+ but when 3198 then there is no `+` because it is end of list, makes sense? when there is more data available show or add `+` at the end of numbers on title when all fetched, no need +.


https://user-images.githubusercontent.com/106739598/213686268-87c9f821-f3c3-4589-a76b-12fa65888534.mp4

feruzm (approved):
lgtm

Issue #1050, opened 2022-09-27
Witness page improvements


feruzm:
Show all witnesses and paginate like voters info

PR #1188, opened 2023-01-09
Witness page improvements

MTSaeed:
**What does this PR?**

Shows search bar at the top of table to search any record.
Provide pagination and sorting. User can apply sorting and pagination.
On page load , it gets the record of 1000 witnesses. But as user reaches in the last page, it fetches the next 1000 records.
When user try to search any witness record, it first checks that search bar witness in the current data. if it does not found in the current record, it will make API call to fetch next 1000 records and then search . This process goes on until all the data fetched and searched.
If search bar witness found in the  current data, it will show that record. if not found, it will show a message of "No results found".
When screen size reaches equal or less than medium breakpoints, it shows the record in row direction and user can scroll infinite until all the records fetched and displayed.

**Steps to reproduce**

1.Type /witnesses after the base URL and you will see the data coming from API.
2. Click on the next page . if you click on the last page, it will take you to the last page and call the API to fetched next 1000 records.
3. Type any thing in the search bar, it will search that record from the current record. if it is not found, it will call API and fetch next 1000 records. This thing goes on until all records fetched. 
4. It will show records in a row direction if screen size equal or less than medium breakPoint and user can perform infinite scrolling until all the records being fetched.

**Issue number** 

fixes #[1050](https://github.com/ecency/ecency-vision/issues/1050)

**Screenshot / Video**

https://user-images.githubusercontent.com/106739598/211297339-46784e16-c053-4d0e-a409-e0e6bb244878.mp4

https://user-images.githubusercontent.com/106739598/211297370-a3a58acc-97ef-4f59-8a6f-54454b61924b.mp4

https://user-images.githubusercontent.com/106739598/211297399-9f1d3268-fe27-4881-bc0e-bbca6dff85f0.mp4


src/common/pages/_witnesses.scss
```diff
@@ -158,3 +162,33 @@
     }
   }
 }
+
+.user-info{
+  margin: 1% 0 0 1%;
```
dkildar, line 167: Do not use percentage please. Use `rem` `em` instead

src/common/pages/_witnesses.scss
```diff
@@ -158,3 +162,33 @@
     }
   }
 }
+
+.user-info{
```
dkildar, line 166: Missed space before `{`

MTSaeed:
Updated.

feruzm:
@MTSaeed couple improvements
- show 30 items on single page
- check mobile view, it is fetching 1000 items and causing too much data to show or load on screen, optimize for mobile devices as well and load as little data as possible so page is more responsive and lean.
- search is using already fetched data, right?

MTSaeed:
> @MTSaeed couple improvements
> 
> * show 30 items on single page
> * check mobile view, it is fetching 1000 items and causing too much data to show or load on screen, optimize for mobile devices as well and load as little data as possible so page is more responsive and lean.
> * search is using already fetched data, right?

@feruzm Its updated now to fetch 300 records per API call. Also displaying 30 items per page now. 

For search, its working just like the proposal votes. When search is performed it searches in the existing data and makes API calls if query is not found in loaded data. Repeats the process of sending API calls and searching until the search results are found.

feruzm:
@MTSaeed if search result is not found, api fetching stops right at the end of fetching or it continuously try to fetch more data even if no new data available?

MTSaeed:
> @MTSaeed if search result is not found, api fetching stops right at the end of fetching or it continuously try to fetch more data even if no new data available?

@feruzm If the last data received is less than the limit. It means there is no more data. It will stop searching and making API calls at this point.

MTSaeed:
> @MTSaeed I don't think we need 300 results on initial fetch, because it makes page load slower we have to manage rendering of 300 items, isn't it? I want page to load as fast as possible and then manage pagination as user desires. For SEO also favorable if page displays, renders asap.

@feruzm do you mean that fetching 300 for web is fine but fetch less data when opened on a mobile device we should reduce this limit. Do you have a number in mind for the limit on mobile devices? One drawback is that we will have to send more API calls for searching if the limit is less.

feruzm:
@MTSaeed that's ok, I don't think people will do more searches outside of 30. I think 30 limit on both web and mobile is ok. If people do search from more lists, we can just show loader and fetching more data to search message somehow. Also they might end up just using pagination which will also fetch more data.

src/common/pages/_witnesses.scss
```diff
@@ -48,23 +48,35 @@
...
   }
+
   .witnesses-table {
     margin-bottom: 20px;
-
+    
     table {
       min-width: 840px;
+
+      @media (max-width: $sm-break) {
```
dkildar, line 68: please, use `@include media-breakpoint-down(sm)`

src/common/pages/_witnesses.scss
```diff
@@ -158,3 +171,36 @@
...
+  margin: 1rem 0 0 1rem;
+}
+
+.table-tools {
+  display: flex;
+  align-items: center;
+  justify-content: space-between;
+  flex-direction: column-reverse;
+
+  @media (max-width: $md-break - 1px) {
```
dkildar, line 185: please, use @include media-breakpoint-down(sm)

src/common/pages/_witnesses.scss
```diff
@@ -158,3 +171,36 @@
...
+  @media (min-width: $lg-break) {
+    flex-direction: row;
+  }
+
+  .sorter {
+    display: flex;
+    align-items: center;
+    margin-bottom: 10px;
+
+    @media (min-width: $lg-break) {
```
dkildar, line 198: please, use `@include media-breakpoint-down(lg)`

MTSaeed:
@dkildar Updated.

dkildar:
LGTM

feruzm (approved):
lgtm

PR #1189, opened 2023-01-11
Voting slider bug

MTSaeed:
**What does this PR?**

Get the previous voted value from the API and set that value to the value of slider.
if user did not voted that post, it check the last voted value of any post from local Storage and set that value of slider.
If there is no last voted value in local Storage, it sets the by default value of slider. (for upSlider: 100 , for dowSlider: -1)

**Steps to reproduce**

1. Open any post and give vote to that post.
2. Open another post and give vote whatever you want to that post.
3. Open the link of that post in another window. you will see it sets the value that you set before while session storage is empty.
4. If the voted value of that post is preset in session storage , it picks the value from session storage and sets the value of slider. if not, then it will get the previous voted value by API call.

**Screenshot / Video**

https://user-images.githubusercontent.com/106739598/212050600-e705e6aa-9e03-4153-b0c0-eee2e40277cb.mp4

### DIFF
--- a/src/common/components/entry-vote-btn/index.tsx
+++ b/src/common/components/entry-vote-btn/index.tsx
@@ -500,23 +500,28 @@ export class EntryVoteBtn extends BaseComponent<Props, State> {
 
   getPreviousVote = async () => {
     const { activeUser, entry } = this.props;
-    const { upVoted } = this.isVoted();
+    const { upVoted, downVoted } = this.isVoted();
     let previousVote;
     if (!activeUser) {
       return null;
     }
 
-    let username = this.props.activeUser?.username! + "-" + this.props.entry.post_id;
-    let type = upVoted ? "up" : "down";
-    let sessValue = ss.get(`vote-value-${type}-${username}`, null);
-    if (sessValue !== null) {
-      return sessValue;
-    }
+    if (upVoted || downVoted) {
+      let username = this.props.activeUser?.username! + "-" + this.props.entry.post_id;
+      let type = upVoted ? "up" : "down";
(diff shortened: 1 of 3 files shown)


MTSaeed:
1- On click of entry vote btn.
    - Before toggeling the dialog fetch the previously voted value:
        * if there is a voted value in local session storage which is the cache. Ftech that. Otherwise fetch vote value from API. This will resolve the problem that if a user votes on a post and then goes to another tab and opens the same profile or feed. His voted value is not displayed as previously all voted values were being fetched from SS.
        * Above step gives us the previously voted value either from SS or from API. If both has no value we get it from LS which has the user's last voted percentage on any post.
2- Send the previous voted value in the EntryDialog. Here getVoteValue function sets the initial values and for upVote and downVote in the initial state. I have modified it in a way that:
    - Check if content was voted and we are getting previously voted value. If yes save that as initial state. If post is never voted take last payout value from LS.
    - We set initialValues and upSliderVal and downSliderVal from above logic. initialValues needs to be set because according to previous logic when slider value is changed and upvoted or downvoted, the system matches the initialValue with the current value. Here our implementation will fix the error that prompts to vote with a different percentage even if the current value was different.

Note: 

The only issue could be that we are fetching the previous vote value first and then toggeling the dialog. As we are taking the values from SS for already voted posts in the current session, the operation is very good in this case. But if the post was voted previously and the value is not in session storage, it cause a very minor delay in opening the dialog. This could be negligible in production as the servers and API speed is fast there.

I was unable to keep this async as the dialog componenet renders first and the vote value is fetched later. This will add some complexity for us to manage.

feruzm (approved):
lgtm

PR #1190, opened 2023-01-13
show Load Draft  button on submit page render

### LINKED ISSUE #1051: Feature: Prompt user to load recent draft upon editor launch
if there is not cached post. like we do in mobile app

MTSaeed:
**What does this PR?**

When submit page renders first time, It checks the key of local_draft from local storage. if the data on that key is empty, it shows the button of "Load Draft".
When user clears the draft, it show the "load Draft button to load the previous saved draft.
When user clicks on "Load Draft" button , it calls the Draft component and display the draft in the form of Modal.
if user is not logged in and try to load the draft, it first ask the user to logged in.

**Steps to reproduce**

1. Open submit page. If it is initially empty, it would show the button of ''Load draft" . If it is not initially empty, it would show the 
    button of ''Save Draft".
2. Click on "Load Draft" button.it first checks the user is logged in or not?. if user is not logged in, it first ask the user to logged 
    in. If user is already logged in, It would show the draft modal to the user to load draft from the recent saved drafts.
3. If user is not logged in and tried to update or save the draft. it first ask the user to logged in and then allow to perform that 
    specific action.

**Issue number** 

Fixes #1051

**Screenshot / Video**

https://user-images.githubusercontent.com/106739598/212295827-437d6e53-388c-4908-a6ef-e8d89113c7f6.mp4

https://user-images.githubusercontent.com/106739598/212296958-c1970da9-db5d-42f2-a534-bb940863d624.mp4


src/common/pages/submit.tsx
```diff
@@ -367,6 +373,20 @@ class SubmitPage extends BaseComponent<Props, State> {
...
   };
 
+  checkDraft = () => {
+    const localDraft = ls.get("local_draft") as PostBase;
+    if (!localDraft) {
+      this.stateSet({ isDraftEmpty: true });
+      return;
+    }
+
+    for (let key in localDraft) {
```
dkildar, line 383: `const`

MTSaeed:
Updated.

src/common/pages/submit.tsx
```diff
@@ -216,7 +216,6 @@ class SubmitPage extends BaseComponent<Props, State> {
 
     this.detectDraft().then();
 
-    this.checkDraft();
```
feruzm: why removing this check?

feruzm: @MTSaeed also, what happens if user doesn't have any drafts stored on backend?

MTSaeed: > why removing this check?

Actually I did the working of this function in the loadLocalDraft function.That's why i think there is no more need to create extra function.

MTSaeed: > @MTSaeed also, what happens if user doesn't have any drafts stored on backend?

It has the current functionality of the Drafts popup. If there are no drafts, the button is still displayed and the drafts pop up opens. It shows the default message for no drafts.

feruzm (approved):
lgtm

PR #1231, opened 2023-02-06
Proposal's title count hotfix

MTSaeed:
**What does this PR?**

It shows the correct number of votes in the proposal's  title.

**Bug**

As you can see in the video attached below.  When user try to gets the votes by just changing the page, it is working fine.
But when user search any thing and then change the page, it shows the wrong count. I debugged the issue and then correct it. 

https://user-images.githubusercontent.com/106739598/216985934-d639e0da-708d-4337-9814-9b7723f0ec93.mp4

**Steps to reproduce**

1. Open any proposal and change the page and see the number of votes on the title.
2. Search any user name and then change the page and jump into the last page. It will show the correct count on that proposal.
3. You can take help from the attached video.

Screenshot / Video 

https://user-images.githubusercontent.com/106739598/216985958-1d225082-23cc-4975-a719-6c4844491e00.mp4

### DIFF
--- a/src/common/components/proposal-votes/index.tsx
+++ b/src/common/components/proposal-votes/index.tsx
@@ -41,6 +41,7 @@ interface Props {
   addAccount: (data: Account) => void;
   onHide: () => void;
   getVotesCount: (votes: number) => void;
+  checkIsMoreData: (check: boolean) => void;
 }
 
 interface State {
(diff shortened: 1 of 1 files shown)


feruzm:
@MTSaeed good find, one more fix... don't show `0+` when no data is fetched yet. Just keep `Votes on ....` without numbers only after fetch is success we can show 1000+ or numbers, makes sense?

MTSaeed:
> @MTSaeed good find, one more fix... don't show `0+` when no data is fetched yet. Just keep `Votes on ....` without numbers only after fetch is success we can show 1000+ or numbers, makes sense?


https://user-images.githubusercontent.com/106739598/217168134-6a4b5b28-aabe-4d07-9cd4-c0754b9f92bb.mp4

feruzm (approved):
lgtm

PR #1236, opened 2023-02-08
Added username validation in sign up page

dkildar:
**What does this PR**
Added username validation in sign up page such as:
1. Min
2. Max
3. First letter in ASCII
4. Regex checking
5. No double hyphens

### DIFF
--- a/src/common/i18n/locales/en-US.json
+++ b/src/common/i18n/locales/en-US.json
@@ -205,7 +205,8 @@
     "username-max-length-error": "Length should be no more than 16 symbols",
     "username-no-ascii-first-letter-error": "First letter should be in ASCII",
     "username-contains-symbols-error": "Should contains letters, numbers, hyphen only",
-    "username-contains-double-hyphens": "Should not contain double hyphens"
+    "username-contains-double-hyphens": "Should not contain double hyphens",
+    "username-exists": "This usernames already exists. Please, try another one"
   },
   "trending-tags": {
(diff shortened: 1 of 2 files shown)


src/common/i18n/locales/en-US.json
```diff
@@ -200,7 +200,12 @@
     "buy-account-li-1": "Instant creation without extra checks and waiting",
     "buy-account-li-2": "3x RC compared to normal new accounts",
     "buy-account-li-3": "300 Points to help you get started",
-    "qr-desc": "Scan with Ecency mobile app to continue"
+    "qr-desc": "Scan with Ecency mobile app to continue",
+    "username-min-length-error": "Length should be at least 3 symbol",
+    "username-max-length-error": "Length should be no more than 16 symbols",
+    "username-no-ascii-first-letter-error": "First letter should be in ASCII",
+    "username-contains-symbols-error": "Should contains letters, numbers, hyphen only",
```
feruzm, line 207: @dkildar, can contain dot as well, only each side of dot should be longer than 3 characters, iirc

feruzm: @dkildar also check these, this is from blockchain source code, much more accurate (apparently this type of username `xyz12` also not supported): https://gitlab.syncad.com/hive/hive/-/blob/master/libraries/protocol/include/hive/protocol/authority.hpp#L76

feruzm (approved):

PR #1243, opened 2023-02-15
Store context

dkildar:
**What does this PR**
Added useMappedStore hook for accessing to store and actions w/o props providing


feruzm (approved):
lgtm, need more tests though

PR #1245, opened 2023-02-18
remove image width and height from source url

### LINKED ISSUE #1232: Animated Thumbnails
The user profile picture animates, which is great, but the post thumbnails do not, when an animated GIF is selected as the thumbnail.

Animated thumbnails help the post stand out when posted into a community.

Screen Shot 2023-02-06 at 11 50 49 AM

calciferlo:
remove image width and height from source url, because constraints for image is size of box and with this property gifs don't work, this fix is according 
Fixes #1232

### DIFF
--- a/src/common/components/entry-list-item/index.tsx
+++ b/src/common/components/entry-list-item/index.tsx
@@ -219,12 +219,12 @@ export default class EntryListItem extends Component<Props, State> {
 
     const imgGrid: string =
       (global.canUseWebp
-        ? catchPostImage(entry, 600, 500, "webp")
-        : catchPostImage(entry, 600, 500)) || noImage;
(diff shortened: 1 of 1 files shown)


feruzm:
@calciferlo thanks for your contribution! Our idea with limiting size of image is to get reasonable image sizes thus reducing data consumption on image downloads. Also image hosted doesn’t resize gif images, so whatever gif size is sent to image proxy server, it will return same size. So we probably need to change something else to display or animate them. I would suggest checking scss files.

feruzm:
@calciferlo closing this, as I think we have fixed this with #1247, update will go out to production in few days. Thank you again for giving us a direction where to look.

PR #1251, opened 2023-02-23
If someone has voted but with zero rshares consider it an upvote

igormuba:
A previous hardfork allowed Resource Credits to be delegated, which in turn allows accounts with zero HP to vote. The result is that an account can have voted posts and the vote is not rendered properly because rshares = 0 due to no HP but having resource credits to vote.

Upvoting works but it is not rendered properly. Downvoting with 0 HP is not possible. So a vote with 0 rshares should be considered an upvote.

Upvoting with zero HP works:
image
image

Downvoting with zero HP is not possible:
image


feruzm (approved):
lgtm

Issue #1157, opened 2022-12-09
Curation history page just like: /@demo/trail


feruzm:
where we show latest voted posts by account and show content as in feed format, it will likely require fetching get_post or get_content of each author/permlink combination returned by get_account_history

igormuba:
@feruzm  just to make sure.

The goal is that, if I enter in someone else's profile, there will be another section
![image](https://user-images.githubusercontent.com/21374091/207714418-0a7c63bd-70ba-46c7-ae1d-81d7dfd269e1.png)

I see every content that said someone else has upvoted or downvoted, right? Bot only votes, no reblogs or anything else, correct?

igormuba:
@feruzm I am working on this on the branch
https://github.com/ecency/ecency-vision/tree/curation-page-history-1157

But I am struggling to understand how blog,posts, comments and replies are fetched so that I can populate the new tab with the correct data

If I manage to understand how that part works it would be a huge advance for me to understand the app, would be a huge progress to figure out all that

igormuba:
Works, but still need to refine by adding:
- Pagination
- If the logged in user has voted/reblogged the loaded posts

https://github.com/ecency/ecency-vision/commit/cca9e3a30dd4a997b442d10841f3b47baa9c96c2

igormuba:
Also: 
If the user of the trail removes the vote, keep the post or remove it from the trail. Will develop considering it should be removed and not shown on the trail.

PR #1255, opened 2023-03-01
Curation page

igormuba:
Issue related to this PR:
https://github.com/ecency/ecency-vision/issues/1157

What it does:
Shows the latest upvotes/downvotes from a user on it's profile, as a curation trail. Paging included.

How to test it:
Go to a user profile, it will show a "trail" tab next to blog and posts
image

It will then show all the posts that the account has upvoted
image


igormuba:
@MTSaeed 

The errors that remain are, from what I understand, the desktop app side, such as `ecency-vision/src/desktop/app/main.dev.ts`

Since this commit did not introduce them I am not sure if I should fix them or leave to avoid unnecessary git conflicts with someone that could be working on that.

The error I could not fix from this PR is the 
```
ERROR in /home/igormuba/ecency-vision/src/common/pages/profile-functional.tsx(290,51):
TS2339: Property 'at' does not exist on type 'BlogEntry[]'.
```
Which seems to be a problem with the Typescript version, but I am not sure, bumping it may solve it.

image

Maybe we could merge it into dev and open a separate issue to fix those errors to make sure someone else is not already doing duplicate work?

Let me know the possible next steps. Thank you.

igormuba:
@feruzm 
You are right, after that the only remaining error is that the typescript version seems to not have `.at`, which is cool, but I have circumvented by just using the traditional `.length-1`

@MTSaeed  now it ruins without errors:
image

feruzm:
@igormuba huh, that's funny `it ruins without errors` 😅

igormuba:
@feruzm it doesn't ruin anymore, it simply runs as expected now 😉

Work from this branch will be useful for the activities page

Let me know if it is good to go or if there is anything else

src/common/api/hive.ts
```diff
@@ -599,6 +601,24 @@ export interface BlogEntry {
...
+  return client.call("condenser_api", "get_account_history", params).then((data) => {
+    let result = data
+      .map((obj: any) => {
+        return { ...obj[1].op[1], num: obj[0] };
+      })
+      .filter((obj: any) => obj.voter === username);
+    return Promise.all(
+      result.map((obj: any) => getPostNew(obj.author, obj.permlink, username, obj.num))
+    );
+  });
```
dkildar, line 619: 1. Need to rewrite as `async/await`;
2. `Promise.all` uses to resolve multiple promises together but in this case you have one promise with array object. Make as in first item and just return your data

src/common/pages/profile-functional.tsx
```diff
@@ -174,7 +189,14 @@ export const Profile = (props: Props) => {
 
       // filter or username changed. fetch posts.
       if (nextSection !== prevMatchSection || `@${nextUsername}` !== prevMatchUsername) {
-        fetchEntries(global.filter, global.tag, false);
+        if (nextSection === "trail") {
+          let data = await getAccountVotesTrail(username.replace("@", ""), -1);
+          //store.dispatch(entriesFETCHEDAct("__trail__", data.reverse(), "", false));
+          setDataTrail({ ...dataTrail, entries: data.reverse() });
+          //setData({ entries: data.reverse(), error: null, hasMore: false, loading: false });
```
dkildar, line 196: Remove this comments please. Its confusing

src/common/api/hive.ts
igormuba: @dkildar  I didn't understand that. 

That function `getAccountVotesTrail` first fetches from the account history what are the posts voted, but still the posts are not resolved, they are in the format 

```
{
  "voter": "igormuba-dev",
  "author": "hispapro",
  "permlink": "soporte-diario-939-publicaciones-valoradas",
  "weight": 10000,
  "num": 142
}
```

So the function maps each of those  to the promise `getPostNew` so each entry in that array is a post promise that will be fetched separately, that is why the `Promise.all` because it is an array of post resolutions, so the return from that function is the posts already resolved

If that doesn't look like a good solution could you elaborate on your suggestion? I couldn't exactly understand what could be improved there. Thanks.

dkildar: You're right. My fault:)

igormuba:
@dkildar @feruzm  pulled from dev, no conflicts, looks good to merge

src/common/api/hive.ts
```diff
@@ -599,6 +601,24 @@ export interface BlogEntry {
 export const getBlogEntries = (username: string, limit: number = dataLimit): Promise<BlogEntry[]> =>
   client.call("condenser_api", "get_blog_entries", [username, 0, limit]);
 
+export const getAccountVotesTrail = (
+  username: string,
+  start = -1,
+  limit = 20
+): Promise<BlogEntry[]> => {
+  let params = [username, start, limit, 1];
+  return client.call("condenser_api", "get_account_history", params).then((data) => {
```
feruzm, line 610: @igormuba you should use `getAccountHistory` api call which is already defined in `common/api/hive.ts` with filters. And you can create bitmask filter for fetching that way, you will only get `vote` operations and much less data to process. For example of `makeBitMaskFilter` check `common/store/transactions/index.ts` line ~105, similarly you can simply create filters for vote operations.

feruzm: @igormuba alright, I have fixed this myself.

dkildar:
Please, remove all `console.log`

igormuba:
@feruzm  @dkildar  maybe it is worth taking a look again? This branch is getting behind often and I need to keep pulling from dev.

src/common/api/hive.ts
```diff
@@ -11,7 +11,7 @@ import { vestsToRshares } from "../helper/vesting";
 import isCommunity from "../helper/is-community";
 
 import SERVERS from "../constants/servers.json";
-import { dataLimit } from "./bridge";
+import { dataLimit, getPost as getPostNew } from "./bridge";
```
MTSaeed, line 14: I think its better to rename the alias getPostNew does not seem intuitive. It tricks us into assuming that there may be a new method for fetching posts. May be give a new intuitive name like getPostsForTrail or getPostsForAccHistory :)

src/common/api/hive.ts
```diff
@@ -586,14 +586,39 @@ export const getSavingsWithdrawFrom = (account: string): Promise<SavingsWithdraw
...
+    start,
+    limit
+  ).then((data) => {
+    let result = data
+      .map((obj: any) => {
+        return { ...obj[1].op[1], num: obj[0] };
+      })
+      .filter((obj: any) => obj.voter === username);
+    return Promise.all(
+      result.map((obj: any) => getPostNew(obj.author, obj.permlink, username, obj.num))
```
MTSaeed, line 617: Please change the variable name obj to some better understandable name. From 612 - 617 obj is used in 3 different functions. May be confusing in the future. Please use different names in all functions if possible

src/common/pages/profile-functional.tsx
```diff
@@ -251,16 +296,42 @@ export const Profile = (props: Props) => {
...
+        ) {
+          const moreData = await getAccountVotesTrail(
+            username.replace("@", ""),
+            data[data.length - 1].num! - 1
+          );
+          data = [...moreData, ...data];
+        }
+        let newDataTrail = _.unionBy(dataTrail.entries, data.reverse(), (obj) => obj.post_id);
+        setDataTrail({ ...dataTrail, entries: newDataTrail, loading: false, fetchedOnce: true });
+      }
```
MTSaeed, line 330: Lines 116 - 136 and 311 - 330 seems to perform almost same tasks. To reduce code duplication and improve readability we can move these lines to one function with some modification to use the function in both cases. Also avoid using generic names like obj

src/common/pages/profile-functional.tsx
```diff
@@ -103,7 +112,31 @@ export const Profile = (props: Props) => {
     const { username, section } = match.params;
     if (!section || (section && Object.keys(ProfileFilter).includes(section))) {
       // fetch posts
-      fetchEntries(global.filter, global.tag, false);
+      if (section === "trail") {
```
MTSaeed, line 115: check comment on 330

src/common/pages/profile-functional.tsx
```diff
@@ -566,16 +645,34 @@ export const Profile = (props: Props) => {
...
-                  if (pinnedEntry) {
-                    entryList.unshift(pinnedEntry);
+                if (
+                  data !== undefined ||
+                  (section === "trail" &&
+                    dataTrail.entries !== undefined &&
+                    dataTrail.entries.length > 0)
+                ) {
+                  let entryList;
+                  switch (section) {
```
MTSaeed, line 655: For this switch case I do not suggest doing this filtering at the time of rendering. May be we check and unshift the pinned posts at the time of fetching enteries. 

One more suggestion is to create a separate component for Trail just like in above lines different components are returned on different sections. This will have following benefits:

- **Maintaining Code Integrity:** As I can see that no data fetching code is present inside Profile component. For profile posts the entries are fetched in the store and stored there. I understand that we cannot save the trail in store as different profiles would be opened by the user and it will eventually make the store heavy. Therefore I suggest that we go to a separate component and do all the data fetching there. 
- **Clean Code:** Above approach would clean the code. Also improve performance as there would be less conditions to check at the time of render. Since the component could be optimized for performance and would be data driver, multiple checks on dataTrail would not be needed at the time of render. 
- **Better State Management:** Fetching data  and state management of dataTrail entries would be simplified and Profile component will retain its current code without adding much complexity to it.

**Challenges:**
- **Handling loading:** Since loading is managed in main (Profile) component. To keep simple we might avoid rendering another loading component in our new component. We might want to use the current loading mechanism from Profile component with our new component. For this we could pass a function from Profile component to the new child component and vice versa to manage loading.
- **Handling bottom reached:** Same approach as the above loading mechanism could be followed. 
There could be other challenges also. I am happy and available to help anytime if needed. :)

We can see that EntryListContent component is returned at last by default here and is still another component. So I think it would be a better approach to create a new component for Trail as well. It will help avoid complexity in Profile component.

Basically I suggest to move the Trail data management outside of the Profile component and in a child component.

@feruzm And @dkildar . Please provide your thoughts on this.

If this approach is followed it might rule out some of my previous comments for the Profile component. But it would still be helpful as @igormuba could consider those suggestions while implementing the new component.

src/common/components/profile-menu/index.tsx
```diff
@@ -55,6 +55,20 @@ export class ProfileMenu extends Component<Props> {
...
     };
+    let secondaryMenu = [
+      {
+        label: _t(`profile.section-trail`),
+        href: `/@${username}/trail`,
+        selected: filter === "trail",
+        id: "trail"
+      }
+    ];
+    const kebabMenuConfig = {
```
MTSaeed, line 66: May be kebab menu is not even needed. IMHO adding a tab in same manner as other tabs would be a better approach. It has following benefits:

- Easy management of responsiveness in tabs and converting them to a dropdown on mobile screens. 
- Highlighting the currently active tabs. Currently Trail is highlighted inside the kebab menu and is not visible if trail is selected. The user does not know if he is viewing the trail page. 

The kebab menu is not responsive. One mobile screens there are two dropdowns, one is kebab menu and other one is the tabs on mobile screens. This does not look suitable. See screenshot for better understanding.

image

MTSaeed:
@igormuba Please read the comment on line 655 before all others. It might alter your approach depending on what you decide after reading it. Also still keep in view other suggestions during implementation if you end up following my suggestion and creating a new component. Feel free to reach for any help.

igormuba:
@MTSaeed I see there were many changes, I am testing what has changed and catching up, for now I am working on two issues that I can see immediately when testing, which is that if a user has not upvoted anything it will load forever (for example http://localhost:3000/@abh12345/trail) 

And the other issue is that if I open a page straight at the trail page it gives an error and crashes

If you have ideas of what to do and what to prioritize I would appreciate, so we can get this feature smoothed out 

Thanks for the feedback and contribution on the PR, by the way! Sincerely appreciate it!

MTSaeed:
> @MTSaeed I see there were many changes, I am testing what has changed and catching up, for now I am working on two issues that I can see immediately when testing, which is that if a user has not upvoted anything it will load forever (for example http://localhost:3000/@abh12345/trail)
> 
> And the other issue is that if I open a page straight at the trail page it gives an error and crashes
> 
> If you have ideas of what to do and what to prioritize I would appreciate, so we can get this feature smoothed out
> 
> Thanks for the feedback and contribution on the PR, by the way! Sincerely appreciate it!

Hi @igormuba Thanks for your sincere remarks. I have fixed these issues in latest commit. Thank you for pointing out.

feruzm (approved):
lgtm, thanks @igormuba and @MTSaeed, good work!

PR #1259, opened 2023-03-07
Refactored styles

dkildar:
**Core changes**
1. Removed duplicated code. Previously `theme-day` and `theme-night` contained same bootstrap styles. It had been done because of overriding bootstrap variables based on themes. I reworked it with CSS-variables(See file: `style.scss` and `theme-*.scss` ). 
2. CSS-variables for theming. Added CSS-variables which switches in `theme-day` and `theme-night` classes.
3. Fixed styles import sequence.

**Improvements**
1. Each loadable page have own `css` file and it will be loaded when page will be appeared. You can build prod and check network for that. It helped to reduce base `style.scss` file.
2. Each component imports style to their `tsx` file. It need for dropping unused styles from `style.scss`. It means that if page use some component then these styles will be added to page styles. E.g. `navbar` uses everywhere and it added to `style.scss` but `entry-item` uses only in profile page and these styles will be in profile page styles file. It will be done automatically on build time. If you make any component loadable then it will have own css file too.

**Breaking changes**
For good working we have to import each component styles to `tsx` not to style. 

**Results**
1. We had a two css files with ~1.8Mb each(theme day and night). It contained whole app styles.
2. Now we have a `style.css` ~ 700Kb(not gzipped). And other styles will be loaded particularly. E.g. if you open main page there will be only `style.css` but if you open communities page there will be two css files: `style.scss` and `communities.css`.


feruzm (approved):
lgtm

PR #1271, opened 2023-03-11
Added caching level for trading view

dkildar:
Added caching level for trading view widget, it works like next:
1. Request data from `useTradingViewApi` hook;
2. It gets time period and bucket size and checks if database have some records in this interval;
3. If data exists in database return them else fetch from API;
4. Save got response to cache based on bucket size and date time.


feruzm (approved):
lgtm

PR #1279, opened 2023-03-16
Feature/login page change

shawnpringle:
Displays Keychani login dialog by default when the keychain extension is installed


src/common/components/login/index.tsx
```diff
@@ -84,6 +90,48 @@ export class LoginKc extends BaseComponent<LoginKcProps, LoginKcState> {
     toggleUIProp("login");
   };
 
+  userSelect = (user: User) => {
```
dkildar, line 93: Please, use `async/await` function

PR #1281, opened 2023-03-17
Show sign up error under button instead of toaster

### LINKED ISSUE #1256: Signup page errors
Human check error shows up quite often even after checking captcha, so there is state setting logic somewhere.

Also move errors them inside form instead of toast messages, so we can give more context.

image

MTSaeed:
**What does this PR?**

Shows Recaptcha error when isVerified state is false.
When user try to register account, it shows error message under the Register button .

**Steps to reproduce**

1.Try to create account and then you will see error message under the "Register" button.As you can take guidelines from the 
    given video.

**Issue number** 

fixes #1256 

**Screenshot / Video**

https://user-images.githubusercontent.com/106739598/225840274-4749e505-77d6-46be-8a35-5ec19b7efec1.mp4

### DIFF
--- a/src/common/pages/sign-up.scss
+++ b/src/common/pages/sign-up.scss
@@ -16,14 +16,15 @@
       display: none;
 
       @media (max-width: $xl-break) {
-        &.register-type, &.buy-account {
+        &.register-type,
+        &.buy-account {
           display: block;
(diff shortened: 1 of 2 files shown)


MTSaeed:
image

PR #1282, opened 2023-03-18
fixed rc balances

Adesojisouljay:


Adesojisouljay:
**What does this PR?**
When there is incoming Rc delegation, it should be reflected in used and available Rc balance.

feruzm:
> **What does this PR?** When there is incoming Rc delegation, it should be reflected in used and available Rc balance.

Share screenshot please

Adesojisouljay:
> 

Screenshot_20230319-104859_Discord

Here it says available 202M, used 1.01T but there is delegation in 4.36Q - 610B which means RC total is more in Quadrillion range, so used amount should be in that range.

feruzm (approved):
lgtm

PR #1290, opened 2023-03-22
brave notifications sounds mute

### LINKED ISSUE #1211: Bug: Brave notifications sound won't mute
Started to get notification alert/sound on Brave browser, even if all notifications are switched off.

MTSaeed:
**What does this PR?**
If notifications are turned off from notification settings the notification won’t pop up and the sound won’t play.
If notification is turned on from settings but notification alerts from preferences are turned off. Only the notification displays but the sound will not play.
If notifications are turned on from settings and notifications alerts from preferences are also turned on then the sound will play and the notification will also display.

**Steps to reproduce**

1. Transfer funds from one account to another account. The source account open in chrome browser and target account opens in brave browser.
2. On the notification alerts setting from preferences and also turn on the transfer setting from notifications. You will hear the sound of notification and also see the notification popup.
3. Turn off the notifications alert from preferences. You will just see the notification popup but the sound won't play.
4. But if you turn off the transfer and any other setting from notifications, you won't see any popup notification or any sound of notifications. 

**Issue number** 

fixes #1211 

**Screenshot / Video**
You can hear the notification sound in the video.

https://user-images.githubusercontent.com/106739598/226862852-bfa89286-56fc-41af-81a3-b849f5032cd0.mp4

### DIFF
--- a/src/common/api/notifications-ws-api.ts
+++ b/src/common/api/notifications-ws-api.ts
@@ -5,7 +5,6 @@ import { ActiveUser } from "../store/active-user/types";
 import defaults from "../constants/defaults.json";
 import { NotifyTypes } from "../enums";
 import { playNotificationSound } from "../util/play-notification-sound";
-import * as ls from "../util/local-storage";
 
 declare var window: Window & {
   nws?: WebSocket;
--- a/src/common/api/notifications-ws-api.ts
+++ b/src/common/api/notifications-ws-api.ts
@@ -5,6 +5,7 @@ import { ActiveUser } from "../store/active-user/types";
 import defaults from "../constants/defaults.json";
 import { NotifyTypes } from "../enums";
 import { playNotificationSound } from "../util/play-notification-sound";
+import * as ls from "../util/local-storage";
 
 declare var window: Window & {
   nws?: WebSocket;
@@ -76,7 +77,10 @@ export class NotificationsWebSocket {
         return;
       }
 
-      await this.playSound();
+      if (ls.get("notifications")) {
+        console.log("IF true");
+        await this.playSound();
+      }
 
       new Notification(_t("notification.popup-title"), { body: msg, icon: logo }).onclick = () => {
         if (!this.hasUiNotifications) {
--- a/src/common/api/notifications-ws-api.ts
+++ b/src/common/api/notifications-ws-api.ts
@@ -77,10 +77,7 @@ export class NotificationsWebSocket {
         return;
       }
 
-      if (ls.get("notifications")) {
-        console.log("IF true");
-        await this.playSound();
-      }
+      await this.playSound();
 
       new Notification(_t("notification.popup-title"), { body: msg, icon: logo }).onclick = () => {
         if (!this.hasUiNotifications) {
--- a/src/common/components/preferences/index.tsx
+++ b/src/common/components/preferences/index.tsx
@@ -13,6 +13,8 @@ import { ActiveUser } from "../../store/active-user/types";
 import { copyContent } from "../../img/svg";
 import * as ls from "../../util/local-storage";
 import "./_index.scss";
+import { useMappedStore } from "../../store/use-mapped-store";
+import { NotifyTypes } from "../../enums/notify-types";
 
 interface Props {
   global: Global;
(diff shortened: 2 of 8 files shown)


feruzm:
@MTSaeed, notification alert is stored on backend and alert (sound + popup) should follow that settings. If vote notification alerts are turned off, then only that should turn off (sound + popup) but other notification types should still popup+sound. 
User profile settings, turning off notification alert should also use same backend api only in this case, it should turn off and on all types of notifications (sount+popup).

feruzm:
@MTSaeed 
Basically these two are slightly different UI but should use same backend api.
Only 1st image you can individually turn on/off notification types.
Screenshot 2023-03-22 at 11 58 08
2nd image you can turn on/off all type of notifications (sound+popup)
Screenshot 2023-03-22 at 11 58 32

MTSaeed:
@feruzm I have noticed that checks for global notification alert setting are not in place. The notification alerts are displayed and sound is played regardless of the global settings. I suggest few things.

feruzm:
@MTSaeed 

- When Settings alert is on/off should reflect notification drawer all off, all on.
- When notification drawer individuals on/off should reflect Settings alert on/off too only it is selective on/off, not all types.

Does this means sense? 

Alternative, expand Settings page dropdown with options like we have on notifications drawer with additional `All` option that can be on/off. That way dropdown is almost same.

MTSaeed:
image

MTSaeed:
@feruzm I have done following implementation.

1- When notifications are turned off from the settings. All notifications are turned off in the notification drawer.
2- If all notifications are turned off, and one of them is turned on in notification drawer. It will turn on the notifications in profile settings. 
3- If all notifications are turned off from the notifications, it will turn off the notifications in the settings. 
4- If all notifications were off previously turning on the notifications from settings will turn on all notifications in the drawer.

MTSaeed:
https://user-images.githubusercontent.com/106739598/230610213-c6bfd661-0221-4223-ae01-5795ca815fde.mp4

src/common/api/notifications-ws-api.ts
```diff
@@ -5,6 +5,7 @@ import { ActiveUser } from "../store/active-user/types";
 import defaults from "../constants/defaults.json";
 import { NotifyTypes } from "../enums";
 import { playNotificationSound } from "../util/play-notification-sound";
+import * as ls from "../util/local-storage";
```
feruzm, line 8: Import in unnecessary

src/common/components/preferences/index.tsx
```diff
@@ -133,7 +147,7 @@ export class Preferences extends BaseComponent<Props, State> {
                 <Form.Label>{_t("preferences.notifications")}</Form.Label>
                 <Form.Control
                   type="text"
-                  value={global.notifications ? 1 : 0}
+                  value={ls.get("notifications") || global.notifications ? 1 : 0}
```
feruzm, line 150: Couldn’t we use api response instead local storage?

src/common/components/user-nav/index.tsx
```diff
@@ -277,7 +279,7 @@ class UserNav extends Component<Props, State> {
                     {unread.toString().length < 3 ? unread : "..."}
                   </span>
                 )}
-                {global.notifications ? bellSvg : bellOffSvg}
+                {ls.get("notifications") || global.notifications ? bellSvg : bellOffSvg}
```
feruzm, line 282: Same here, api response be stored in redux and accessed instead of local storage which is sometimes unreliable

src/common/api/notifications-ws-api.ts
MTSaeed: @feruzm 
Yes, Removed.

src/common/components/preferences/index.tsx
MTSaeed: yes.
Instead of updating the localStorage and then use the value of localStorage, i updated the store state and then use it.

src/common/components/user-nav/index.tsx
MTSaeed: Resolved.

MTSaeed:
image

feruzm (approved):
lgtm

PR #1302, opened 2023-04-04
Account recovery expand

MTSaeed:
**What does this PR?**

shows the warning message if the account username and current recovery agent name is same.
shows the input field for email if user try to set ecency as its recovery agent.
Shows the latest email in the email field if user has already set ececny its recovery agent. 
If user change its recovery agent from ececny to any other account, delete the last set email from database. Before deleting from the database ,check if it is only one email in database then not delete from the database. if its not the last email in the database then delete the latest one.

https://www.loom.com/share/842c8a2ca60f4737af43b9615b0408fe

 
**Steps to reproduce**

1. Set ececny as your recovery agent and you will see an input field appear to ask to enter email for verification purpose.
2. You can take more guidance from the attached video.

**Screenshot/Video**


MTSaeed:
image

MTSaeed:
https://user-images.githubusercontent.com/106739598/229779063-322aaa47-e75d-4c52-98d6-d4284d9cf0c7.mp4

src/common/api/private-api.ts
```diff
@@ -591,3 +589,31 @@ export const getAnnouncementsData = async (): Promise<Announcement[]> => {
     throw error;
   }
 };
+export interface Recoveries {
+  username: string;
+  email: string;
+  public_keys: Object;
```
dkildar, line 595: Let's clarify it with Record: `Record<string, ?>`. Replace `?` with your type

src/common/api/private-api.ts
```diff
@@ -591,3 +589,31 @@ export const getAnnouncementsData = async (): Promise<Announcement[]> => {
...
+}
+
+export interface getRecoveriesEmail extends Recoveries {
```
dkildar, line 598: Interfaces/classes/consts shouldn't be named as action. Use nouns instead: `GetRecoveriesEmailResponse`

src/common/api/private-api.ts
```diff
@@ -591,3 +589,31 @@ export const getAnnouncementsData = async (): Promise<Announcement[]> => {
...
+
+export const getRecoveries = (username: string): Promise<getRecoveriesEmail[]> => {
+  const data = { code: getAccessToken(username) };
+  return axios.post(apiBase(`/private-api/recoveries`), data).then((resp) => resp.data);
+};
+
+export const addRecoveries = (
+  username: string,
+  email: string,
+  public_keys: Object
```
dkildar, line 610: I would like to ask you use `camelCase`

src/common/components/recovery-account/index.tsx
```diff
@@ -23,33 +23,59 @@ import {
...
+
 export default function AccountRecovery(props: Props) {
   const [keyDialog, setKeyDialog] = useState(false);
   const [inProgress, setInProgress] = useState(false);
   const [disabled, setDisabled] = useState(true);
+  const [isEcency, setIsEcency] = useState(false);
   const [popOver, setPopOver] = useState(false);
   const [step, setStep] = useState(1);
   const [toError, setToError] = useState("");
+  const [accountData, setAccountData] = useState<FullAccount>();
```
dkildar, line 49: Let's make `FullAccount | undefined` because you don't have initial value

src/common/components/recovery-account/index.tsx
```diff
@@ -23,33 +23,59 @@ import {
...
+  const [recoveryEmail, setRecoveryEmail] = useState("");
+  const [toWarning, setToWarning] = useState("");
+  const [recoveryEmails, setRecoveryEmails] = useState<getRecoveriesEmail[]>([]);
   const [currRecoveryAccount, setCurrRecoveryAccount] = useState("");
   const [newRecoveryAccount, setNewCurrRecoveryAccount] = useState("");
   const [pendingRecoveryAccount, setPendingRecoveryAccount] = useState("");
 
+  const ECENCY = "ecency";
```
dkildar, line 57: Please, move it outside of component declaration because this property will be created each render time

src/common/components/recovery-account/index.tsx
```diff
@@ -75,10 +101,20 @@ export default function AccountRecovery(props: Props) {
       setToError("");
       return;
     } else {
+      if (e.target.value === ECENCY) {
+        setIsEcency(true);
+      } else {
+        setIsEcency(false);
+      }
```
dkildar, line 108: Lets simplify it: `setIsEcency(e.target.vaue === ECENCY)`

src/common/components/recovery-account/index.tsx
```diff
@@ -75,10 +101,20 @@ export default function AccountRecovery(props: Props) {
...
+
       const resp = await getAccount(e.target.value);
       if (resp) {
-        setDisabled(false);
-        setToError("");
+        if (e.target.value === ECENCY) {
+          setDisabled(true);
+        } else {
+          setDisabled(false);
+          setToError("");
```
dkildar, line 116: Same here: `setDisabled(e.target.value === ECENCY)`

src/common/components/recovery-account/index.tsx
```diff
@@ -99,9 +135,23 @@ export default function AccountRecovery(props: Props) {
...
 
+  const handleIsEcency = async () => {
+    if (isEcency) {
+      let response = await addRecoveries(props.activeUser?.username!, recoveryEmail, {
+        publick_keys: accountData!.owner.key_auths[0]
+      });
+      setIsEcency(false);
+    } else {
+      if (recoveryEmails.length >= 2) {
+        let response = await deleteRecoveries(props.activeUser?.username!, recoveryEmails[0]._id);
```
dkildar, line 146: And what next with this response? If it won't be used just remove it:)

src/common/components/recovery-account/index.tsx
```diff
@@ -147,6 +199,16 @@ export default function AccountRecovery(props: Props) {
...
   };
 
+  const handleRecoveryEmail = (e: React.ChangeEvent<typeof FormControl & HTMLInputElement>) => {
+    setRecoveryEmail(e.target.value);
+    const emailRegex = /^\w+([\.-]?\w+)*@\w+([\.-]?\w+)*(\.\w{2,3})+$/;
+    if (emailRegex.test(e.target.value)) {
+      setDisabled(false);
+    } else {
+      setDisabled(true);
+    }
```
dkildar, line 209: `setDisabled(!emailRegex.test(e.target.value)`

MTSaeed:
@dkildar and @feruzm 
All comments have been resolved.

feruzm:
@MTSaeed If recovery agent is already set to ecency and user just updates email, then we don't need to do blockchain transaction, we could just do backend api call to update/add the email. From your latest video, I see that blockchain transaction still being performed, is that correct?

MTSaeed:
> @MTSaeed If recovery agent is already set to ecency and user just updates email, then we don't need to do blockchain transaction, we could just do backend api call to update/add the email. From your latest video, I see that blockchain transaction still being performed, is that correct?


@feruzm yes, Updated.

https://user-images.githubusercontent.com/106739598/231684189-7bf4d28b-938d-4482-a872-8de81468faac.mp4

src/common/components/recovery-account/index.tsx
```diff
@@ -65,12 +65,12 @@ export default function AccountRecovery(props: Props) {
...
     setAccountData(account);
     const { recovery_account } = account;
-    setCurrRecoveryAccount(recovery_account);
+    setCurrRecoveryAccount(ECENCY);
     if (recovery_account === props.activeUser?.username) {
       setToWarning(_t("account-recovery.same-recover-agent-suggestion"));
     }
 
-    if (recovery_account === ECENCY) {
+    if (ECENCY === ECENCY) {
```
feruzm, line 73: @MTSaeed this condition always true, why needed?

MTSaeed: @feruzm My mistake😬
I set this condition to test for the case of ECENCY and after that I forgot to reset that condition. Now i update the condition.
Thanks!

feruzm: @MTSaeed I just tried, it is sending email to backend but not changing recovery agent. Also when I type into new agent field, it sometimes says account doesn't exist error. You should make sure email update only happens if I set new agent to current agent, that way, it should only change email without blockchain broadcast, otherwise, it should broadcast to blockchain as well as send email change to backend with keys.

feruzm: Also please pull before fix above issue, as I pushed couple minor improvements.

MTSaeed: Hi @feruzm 
I have done some changes in the Logic. I am going to explain the use cases. If you feel to change in any use case, please mention the use case number too.

1- If the current Recovery agent is any account other than ECENCY and you want to set new recover agent other than ECENCY, it would perform the broadcast transaction.

2- If the current Recovery agent is any account other than ECENCY and you want to set ECENCY as your new recover agent, it would send API call to add email in the database and as well as it will perform broadcast transaction to change recover agent to ECENCY.

3- If the current recovery agent is ECENCY and you just want to update the email, it will allow you to update the email and it will not show you the PopOver and also it will not perform broadcast transaction. it will just update the email by sending API call to the Database.

4- If the current recovery agent is ECENCY and user want to set any account other than ECENCY to their new recover agent, it will perform broadcast transaction to update the new recover agent and also send the API call to delete the recent email.

MTSaeed: @feruzm This is just the video of use case 1 & 2 when the current recover agent is other than ecency.

https://user-images.githubusercontent.com/106739598/232410999-e3d26570-8af7-4122-99ae-20d0aec0c4a7.mp4

feruzm: @MTSaeed 4. Do not delete email from backend, allow user to change recovery agent broadcasting transaction, but no api call needed.

MTSaeed: > @MTSaeed 4. Do not delete email from backend, allow user to change recovery agent broadcasting transaction, but no api call needed.

Updated. Now if user try to change recover agent from ECENCY to any other account, it will not send API call to delete the email from database. It will just broadcast the transaction.

feruzm: @MTSaeed consider using Debounce for account typing, so we don't do multiple calls to rpc server unnecessarily. debounce 500 ms should be sufficient.

src/common/components/recovery-account/index.tsx
```diff
@@ -66,23 +65,21 @@ export default function AccountRecovery(props: Props) {
...
-  };
-
-  const getRecoveryRequest = async () => {
     const resp = await findAccountRecoveryRequest(props.activeUser!.username);
     if (resp.requests.length) {
       setPendingRecoveryAccount(resp.requests[0].recovery_account);
       setPopOver(true);
     }
+
+    if (ECENCY === ECENCY) {
```
dkildar: It always `true` 🤔

src/common/components/recovery-account/index.tsx
MTSaeed: yes. updated.

feruzm:
@MTSaeed added proper debouncing, please check for future reference

feruzm (approved):
lgtm

PR #1352, opened 2023-06-01
check submit input time out

Adesojisouljay:
### What does this PR? 
A simple fix on submit body input as shown in the screen capture.

@feruzm please check if this fix for this bug is okay, for now I think removing set time out works, but we might want to consider if there is a more suitable solution. If this works well, we can clean those comments in code and merge.

**Screenshot**

https://github.com/ecency/ecency-vision/assets/88087890/1b04e12a-b1e7-4a7b-b657-03ef2e6a286e


feruzm (approved):
lgtm

PR #1370, opened 2023-06-17
Get rid of duplicated requests

dkildar:


feruzm (approved):
lgtm, will continue testing and report if new issues are discovered

PR #1447, opened 2023-08-17
Emojis 2.0

dkildar:


feruzm (approved):
lgtm, will do more tests on alpha.

PR #1450, opened 2023-08-22
Feat/usernav to funtional

Adesojisouljay:
### What does this PR?
There is no change in terms of interface/UI. 
UserNav is just being converted from class to functional component


feruzm (approved):
lgtm

PR #1455, opened 2023-08-29
Submit 2.1: Added body manager

dkildar:
Added body manager for controlling video uploading metadata while user make changes in the editor


feruzm (approved):
lgtm

PR #1461, opened 2023-09-10
Bugfix/app crash

dkildar:


feruzm (approved):
lgtm, for video types we should probably add `video/webm` as well

PR #1562, opened 2024-02-27
Fixed sidebar error when open it immediately after page load

dkildar:


src/common/components/decks/deck-toolbar/deck-toolbar.tsx
```diff
@@ -89,7 +89,6 @@ export const DeckToolbar = ({ isExpanded, setIsExpanded, history }: Props) => {
           setIsExpanded={setIsExpanded}
         />
         <DeckToolbarBaseActions
-          history={history}
```
feruzm, line 92: @dkildar I think, this is needed as component still has history prop

feruzm (approved):

PR #1573, opened 2024-03-12
Chats: added contacts pinning

dkildar:


feruzm (changes requested):
@dkildar once (un)pinned we should probably show toast message saying contact is (un)pinned. Also on mobile we should consider adding swipe left to see options from contact list or tap and hold to see options.

feruzm (approved):
lgtm

PR #1608, opened 2024-05-14
Fixed emoji picker wrong selection

dkildar:


feruzm (approved):
interesting, that was the main reason?! let's try

